PR #1, opened 2018-03-12
turn some warnings into errors to avoid spamming the log

mzanetti:


mzanetti:
Jenkins, land this please

PR #2, opened 2018-03-12
Texas instruments sensortag

mzanetti:


mzanetti:
TODOs:
- Need to be able to disable certain sensors from the config to reduce battery consumption
- Move configurations to be device configs, not plugin configs, at least as far as possible
- make motion detection implement the alert interface
- Sensor values seem to come in too often and spam the system. Perhaps tune them down a bit? make it configurable?

PR #6, opened 2018-03-13
don't try to ping on ipv6 link local address, that won't work

mzanetti:


mzanetti:
Jenkins, land this please

PR #17, opened 2018-03-14
added dweet.io plug-in

Boernsman:


mzanetti:
Jenkins, land this please

dweetio/deviceplugindweetio.cpp
```diff
@@ -0,0 +1,265 @@
...
+        Device *device = m_postReplies.value(reply);
+        m_postReplies.remove(reply);
+        ActionId actionId = m_asyncActions.value(reply);
+
+        // check HTTP status code
+        if ((status != 200) && (status != 204)) {
+            qCWarning(dcDweetio) << "Update reply HTTP error:" << status << reply->errorString() << reply->readAll();
+            emit actionExecutionFinished(actionId, DeviceManager::DeviceErrorMissingParameter);
+            setConnectionStatus(false, device);
+            reply->deleteLater();
```
mzanetti: hint: you could call reply->deleteLater in the beginning of the method so you don't have to take care of it in every return path. it will only be deleted once the function is finished (actually only when the event loop idles again next time) so you can still work with it, even after calling deleteLater() within the same function.

mzanetti: I've rebased it on master and fixed the merge conflict myself.

mzanetti:
Jenkins, test this please

mzanetti:
Jenkins, test this please

mzanetti:
Jenkins, land this please

PR #20, opened 2018-03-16
Fix avea plugin

t-mon:


t-mon:
Jenkins, land this please

PR #27, opened 2018-03-26
drop logic to mark hue as unreachable if it doesn't reply in time.

mzanetti:
instead, always wait for the reply and mark it as gone if the
reply is bad


t-mon:
Jenkins, land this please

PR #35, opened 2018-04-27
Force update available to false if an update is running

t-mon:


mzanetti:
Jenkins, land this please

PR #36, opened 2018-04-30
don't cache snap update available state

mzanetti:


mzanetti:
Jenkins, land this please

PR #40, opened 2018-06-07
make use of interfaces in netatmo plugin

mzanetti:


mzanetti:
Jenkins, land this please

PR #42, opened 2018-06-13
dedupe scan targets for networkdetector discovery

mzanetti:


mzanetti:
Jenkins, land this please

PR #43, opened 2018-06-14
trim pushbullet token during setup

mzanetti:


mzanetti:
Jenkins, land this please

PR #45, opened 2018-06-25
fix a crash when a device is remove which never registered a timer

mzanetti:


mzanetti:
Jenkins, land this please

PR #75, opened 2019-01-12
New Plugin: Serialport commander

Boernsman:
New plug-in to send and receive strings over a serial port


serialportcommander/devicepluginserialportcommander.cpp
```diff
@@ -0,0 +1,229 @@
...
+            //device already in use
+            qCDebug(dcSerialPortCommander()) << "Found serial port that is already used:" << port.portName();
+        } else {
+            //Serial port is not yet used, create now a new one
+            qCDebug(dcSerialPortCommander()) << "Found serial port:" << port.portName();
+            QString description = port.manufacturer() + " " + port.description();
+            DeviceDescriptor descriptor(deviceClassId, port.portName(), description);
+            ParamList parameters;
+
+            if (deviceClassId == serialPortCommanderDeviceClassId) {
```
mzanetti: This if seems really at the wrong place...
a) it will never be false given there's only one device class (unless there's a bug in the core)
b) even if a second deviceclass is introduced, this will will still emit devicesDiscovered but with broken parameters, so it is still not be the wanted behavior... I would just remove it, but if you think it's better to have this check, it should probably be around the loop in line 117

serialportcommander/devicepluginserialportcommander.cpp
```diff
@@ -0,0 +1,229 @@
...
+
+
+DeviceManager::DeviceError DevicePluginSerialPortCommander::discoverDevices(const DeviceClassId &deviceClassId, const ParamList &params)
+{
+    Q_UNUSED(params)
+    // Create the list of available serial interfaces
+    QList<DeviceDescriptor> deviceDescriptors;
+
+    Q_FOREACH(QSerialPortInfo port, QSerialPortInfo::availablePorts()) {
+        if (m_usedInterfaces.contains(port.portName())){
```
mzanetti: I cannot find any place where you add someting to m_usedInterfaces...

serialportcommander/devicepluginserialportcommander.cpp
```diff
@@ -0,0 +1,229 @@
...
+    QString interface;
+    QSerialPort *serialPort;
+
+    if (device->deviceClassId() == serialPortCommanderDeviceClassId) {
+
+        m_usedInterfaces.removeAll(device->paramValue(serialPortCommanderDeviceSerialPortParamTypeId).toString());
+        serialPort = m_serialPorts.value(device);
+        serialPort->close();
+        m_serialPorts.remove(device);
+        serialPort->deleteLater();
```
mzanetti: hint:

you can do
```
QSerialPort *serialPort = m_serialPorts.take(device);
serialPort->close();
serialPort->deleteLater();
```

serialportcommander/devicepluginserialportcommander.cpp
```diff
@@ -0,0 +1,229 @@
...
+    }
+    qDebug(dcSerialPortCommander()) << "Message received" << data;
+
+    device->setStateValue(serialPortCommanderInputDataStateTypeId, data);
+    emitEvent(Event(serialPortCommanderTriggeredEventTypeId, device->id()));
+}
+
+void DevicePluginSerialPortCommander::onSerialError(QSerialPort::SerialPortError error)
+{
+    Q_UNUSED(error)
```
mzanetti: please add a
`qCWarning(dcSerialPortCommander) << "Serial Port error happened:" << error;`
or similar

serialportcommander/devicepluginserialportcommander.h
```diff
@@ -0,0 +1,65 @@
...
+
+public:
+    explicit DevicePluginSerialPortCommander();
+
+    DeviceManager::DeviceSetupStatus setupDevice(Device *device) override;
+    void deviceRemoved(Device *device) override;
+    DeviceManager::DeviceError discoverDevices(const DeviceClassId &deviceClassId, const ParamList &params);
+
+    DeviceManager::DeviceError executeAction(Device *device, const Action &action) override;
+    void init() override;
```
mzanetti: Can drop this

serialportcommander/devicepluginserialportcommander.json
```diff
@@ -0,0 +1,118 @@
...
+                                    "type": "QString",
+                                    "inputType": "TextArea"
+                                }
+                            ]
+                        }
+                    ],
+                    "stateTypes": [
+                        {
+                            "id": "b98fdacc-59d7-41c4-b790-1fdca50dfb22",
+                            "name": "inputData",
```
mzanetti: Note that this will be misleading in the case where the same inputData is received a second time because it will not trigger the stateChanged event!
Are you sure you want to offer this? What is your use case?

mzanetti:
Oh, and another note:

you probably want to add a param with the input data to the triggered event. :)

debian/control
```diff
@@ -748,6 +748,20 @@ Description: nymea.io plugin for remote ssh connection
  .
  This package will install the nymea.io plugin for unipi devices
 
+Package: nymea-plugin-serialportcommander
+Architecture: any
+Depends: ${shlibs:Depends},
+         ${misc:Depends},
+         nymea-plugins-translations,
```
t-mon: `libqt5serialport5`

PR #79, opened 2019-01-24
EQ-3: Add support for the Eqiva Bluetooth thermostat

mzanetti:


eq-3/eq-3.pro
```diff
@@ -1,8 +1,8 @@
 include(../plugins.pri)
 
-QT += network
+QT += network bluetooth
 
-TARGET = $$qtLibraryTarget(nymea_deviceplugineq3)
+TARGET = $$qtLibraryTarget(nymea_deviceplugineq-3)
```
t-mon: you probably need to update also the debian install script :)

PR #90, opened 2019-02-17
WIP Keba: added interfaces

Boernsman:
* Reworked Keba plug-in to make use of interfaces.
* Added discovery setup method


keba/devicepluginkeba.cpp
```diff
@@ -30,119 +31,131 @@
...
     }
 
-    QHostAddress address = QHostAddress(device->paramValue(wallboxDeviceIpParamTypeId).toString());
+    qCDebug(dcKebaKeContact()) << "Setting up a new device:" << device->name() << device->paramValue(wallboxDeviceIpAddressParamTypeId).toString();
+    QHostAddress address = QHostAddress(device->paramValue(wallboxDeviceIpAddressParamTypeId).toString());
 
     //Check if the IP is empty
     if (address.isNull()) {
-        delete m_kebaSocket;
+        qCWarning(dcKebaKeContact()) << "Address is empty";
```
mzanetti: why did the delete go away? I think this is still needed.

What would probably be even better is if you would move the check for if(address.isNull()) at the very top of the setupDevice method and only start allocating memory once you are sure the incoming params are ok.

mzanetti: Also, in case this fails the timer stays allocated and ticks away... So yeah, this method needs to be reworked to properly handle error cases.

keba/devicepluginkeba.cpp
```diff
@@ -30,119 +31,131 @@
...
 
-        if(action.actionTypeId() == wallboxMaxCurrentActionTypeId){
-            // Print information that we are executing now the update action
-            qCDebug(dcKebaKeContact()) << "Update max current to : " << action.param(wallboxMaxCurrentActionMaxCurrentParamTypeId).value().toString();
-            QByteArray datagram;
-            datagram.append("curr " + QVariant(action.param(wallboxMaxCurrentActionMaxCurrentParamTypeId).value().toInt()*1000).toString());
-            qCDebug(dcKebaKeContact()) << "Datagram : " << datagram;
-            m_kebaSocket->writeDatagram(datagram.data(),datagram.size(), QHostAddress(device->paramValue(wallboxDeviceIpParamTypeId).toString()) , 7090);
+            int current = action.param(wallboxMaxChargingCurrentActionMaxChargingCurrentParamTypeId).value().toInt();
+            kebaConnection->setMaxAmpere(current);
```
mzanetti: If you are sure the action succeeded here, you should also immediately set
`device->setStateValue(wallboxMaxChargingCurrentStateTypeId, current);`
so that the UI reflects new values right away and doesn't have to wait for the next time we poll the state.

keba/devicepluginkeba.cpp
```diff
@@ -30,119 +31,131 @@
...
-            }
-            else{
-                datagram.append("ena 0");
-            }
-            qCDebug(dcKebaKeContact()) << "Datagram : " << datagram;
-            m_kebaSocket->writeDatagram(datagram.data(),datagram.size(), QHostAddress(device->paramValue(wallboxDeviceIpParamTypeId).toString()) , 7090);
+
+        if(action.actionTypeId() == wallboxPowerActionTypeId) {
+            bool state = action.param(wallboxPowerActionPowerParamTypeId).value().toBool();
+            kebaConnection->enableOutput(state);
```
mzanetti: Same as above

keba/devicepluginkeba.cpp
```diff
@@ -153,91 +166,158 @@ void DevicePluginKeba::readPendingDatagrams()
...
+        foreach (Device *device, myDevices()) {
+            if (device->paramValue(wallboxDeviceIpAddressParamTypeId).toString() == sender.toString())
+            {
+                senderUnkown = false;
+                KebaConnection *kebaConnection = m_kebaConnections.value(device);
+                if (!kebaConnection) {
+                    qCWarning(dcKebaKeContact()) << "No Keba Connection";
+                    continue;
+                }
+                kebaConnection->onAnswerReceived();
```
mzanetti: Tbh I find this a bit odd... The talking to the wallbox is split across 3 classes, KebaConnection, KebaSocket and the main plugin class... Wouldn't it be better if KebaSocket would be a private member of KebaConnection and all communication would go through KebaConnection, keeping the KebaSocket invisble to the outside of KebaConnection? 

What bugs me here is that sending of data is composed in KebaConnection, but parsing of received data is done in the main plugin class... Afaict, this onAnswerReceived() is done in order to inform the connection that it is ready to send the next chunk of data... This really should not be needed at all with a cleaner architecture.

keba/devicepluginkeba.cpp
```diff
@@ -153,91 +166,158 @@ void DevicePluginKeba::readPendingDatagrams()
...
+                        }
+                        else if(data.value("Plug").toString() == "3"){
+                            device->setStateValue(wallboxPlugStateStateTypeId,"Locked plug on charging station");
+                        }
+                        else if(data.value("Plug").toString() == "5"){
+                            device->setStateValue(wallboxPlugStateStateTypeId,"Plugged on charging station and vehicle");
+                        }
+                        else if(data.value("Plug").toString() == "7"){
+                            device->setStateValue(wallboxPlugStateStateTypeId,"Locked plug on charging station and vehicle");
+                        }
```
mzanetti: Would be nicer as

```
bool ok;
int plugId = data.value("Plug").toInt(&ok);
if (!ok) { //Error handling }
switch (plugId) {
  case 1:
  ...
}
```

keba/devicepluginkeba.cpp
```diff
@@ -153,91 +166,158 @@ void DevicePluginKeba::readPendingDatagrams()
...
-            datagram.append("report 3");
-            qCDebug(dcKebaKeContact()) << "datagram : " << datagram;
-            socket->writeDatagram(datagram.data(),datagram.size(), QHostAddress(m_kebaDevices.value(sender)->paramValue(wallboxDeviceIpParamTypeId).toString()) , 7090);
         }
-        else if(data.value("ID").toString() == "3"){
-            //power of current charging session
-            m_kebaDevices.value(sender)->setStateValue(wallboxPowerStateTypeId,data.value("E pres").toInt() / 1000);
-            //current phase 1
-            m_kebaDevices.value(sender)->setStateValue(wallboxCurrentStateTypeId,data.value("I1").toInt() * 1000);
+        if (host.hostName().contains("Keba") && !deviceAddressUsed) {
```
mzanetti: In order to allow reconfiguring the device, you should also create a DeviceDescriptor even if deviceAddressUsed is true, but set the DeviceDescriptor.deviceId to the existing device's id.

mzanetti: (This is new functionality in nymea since 0.11)

keba/devicepluginkeba.cpp
mzanetti: If you're not sure about the success yet, you should not emit DeviceErrorNoError in the next line but use DeviceErrorAsync instead.

PR #91, opened 2019-03-04
Netatmo: Update API changes and improve refresh behavior

t-mon:


netatmo/devicepluginnetatmo.cpp
```diff
@@ -113,6 +117,12 @@ void DevicePluginNetatmo::deviceRemoved(Device *device)
         OAuth2 * authentication = m_authentications.key(device);
         m_authentications.remove(authentication);
         authentication->deleteLater();
+
+        if (m_pluginTimer) {
+            m_pluginTimer->deleteLater();
+            m_pluginTimer = nullptr;
+        }
+
```
mzanetti: Needs to check if there are other netatmos still left

PR #103, opened 2019-05-10
Removed plugins: Orderbutton, Plantcare and ws2812

Boernsman:
remove plug-ins that have been very use case specific and used only once.


debian/control
```diff
@@ -829,9 +781,6 @@ Package: nymea-plugins-merkurboard
 Section: libs
 Architecture: all
 Depends: nymea-plugin-osdomotics,
```
t-mon: You should probably remove also the `nymea-plugin-osdomotics` meta package since that is not required any more.

Boernsman: I was not sure about that, thought the nymea-plugin-osdomotics plug-in is more generic and might be useful for the IoT Austria foundation

PR #106, opened 2019-05-14
Fix serialportcommander

Boernsman:


serialportcommander/devicepluginserialportcommander.h
```diff
@@ -40,23 +42,22 @@ class DevicePluginSerialPortCommander : public DevicePlugin
...
 private:
+    QTimer *m_reconnectTimer = nullptr;
     QHash<Device *, QSerialPort *> m_serialPorts;
-    QList<QString> m_usedInterfaces;
 
 private slots:
     void onReadyRead();
-    void onSerialError(QSerialPort::SerialPortError error);
-    void onBaudRateChanged(qint32 baudRate, QSerialPort::Direction direction);
+    void onSerialError(const QSerialPort::SerialPortError &error);
```
t-mon: you don't need the const & for enums

PR #108, opened 2019-05-15
New Plugin: CoinMarketCap

Boernsman:


coinmarketcap/deviceplugincoinmarketcap.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+
+    // check JSON file
+    QJsonParseError error;
+    QJsonDocument jsonResponse = QJsonDocument::fromJson(reply->readAll(), &error);
+    reply->deleteLater();
+
+    if (error.error != QJsonParseError::NoError) {
+        qCWarning(dcCoinMarketCap()) << "Update reply JSON error:" << error.errorString();
+        reply->deleteLater();
```
mzanetti: You're calling `reply->deleteLater()` a second time here... See previous comment and just drop this one.

coinmarketcap/deviceplugincoinmarketcap.cpp
```diff
@@ -0,0 +1,184 @@
...
+    foreach (Device *device, myDevices()) {
+        if (device->deviceClassId() == currentPricesDeviceClassId) {
+            getPriceCall(device);
+        }
+    }
+}
+
+void DevicePluginCoinMarketCap::onPriceCallFinished()
+{
+    QNetworkReply *reply = static_cast<QNetworkReply *>(sender());
```
mzanetti: Just add a `reply->deleteLater();` after this and drop all the other ones below. You want to call that in any case and you don't want to risk forgetting it in some cases (or calling it twice as it happens here)

mzanetti:
Needs conflict resolving.

mzanetti:
Jenkins, retest this please

coinmarketcap/README.md
```diff
@@ -0,0 +1,4 @@
+#CoinMarketCap
+
+The coin market cap plugin gets the latest crypto prices from coin market cap and displays it in your favorite fiat currency.
+
```
mzanetti: Mind listing which crypto currencies it supports? And maybe some more info about it, dunno, perhaps config options, where it gets the data from (if that's of any interest to a user) etc.

coinmarketcap/translations/de_DE.ts
```diff
@@ -0,0 +1,175 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mzanetti: filename is wrong, needs to be pluginid-de_DE.ts, or probably even better just "pluginid-de.ts" in order to avoid the de_AT issue we had in the app.

coinmarketcap/translations/en_US.ts
```diff
@@ -0,0 +1,175 @@
+<?xml version="1.0" encoding="utf-8"?>
```
mzanetti: same, pluginid missing in filename

debian/control
```diff
@@ -91,7 +91,8 @@ Depends: ${shlibs:Depends},
          ${misc:Depends},
          nymea-plugins-translations,
 Description: nymea.io plugin for bose soundtouch
- The nymea daemon is a plugin based IoT (Internet of Things) server. The
+
+The nymea daemon is a plugin based IoT (Internet of Things) server. The
```
mzanetti: is this wanted? doesn't it break the file format if there's an empty line here?

mzanetti: Indeed, fails the build: 
dpkg-buildpackage: error: syntax error in debian/control at line 95: line with unknown format (not field-colon-value)

mzanetti:
merged

PR #111, opened 2019-05-23
UniPi: Removed dependency on evok

mzanetti:


unipi/unipi.pro
```diff
@@ -2,12 +2,32 @@ include(../plugins.pri)
...
+    neuron.h \
+    neuronextension.h \
+    mcp23008.h \
+    i2cport.h \
+    unipi.h \
+    i2cport_p.h \
+    mcp3422.h
 
+MAP_FILES.files = modbus_maps/
+MAP_FILES.path = /usr/share/nymea/modbus/
```
mzanetti: this won't build in yocto... You should install to [QT_INSTALL_PREFIX]/share/nymea/modbus/

unipi/neuron.cpp
```diff
@@ -0,0 +1,655 @@
...
+
+
+bool Neuron::loadModbusMap()
+{
+    QStringList fileCoilList;
+    QStringList fileRegisterList;
+
+    switch (m_neuronType) {
+    case NeuronTypes::S103:
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_S103/Neuron_S103-Coils-group-1.csv"));
```
mzanetti: instead of hardcoding the path, use QStandardPaths::standardLocation(QStandardPaths::GenericDataLocation).first() instead

unipi/neuron.cpp
```diff
@@ -0,0 +1,655 @@
...
+        break;
+    case NeuronTypes::L503:
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L503/Neuron_L503-Coils-group-1.csv"));
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L503/Neuron_L503-Coils-group-2.csv"));
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L503/Neuron_L503-Coils-group-3.csv"));
+        break;
+    case NeuronTypes::L513:
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L513/Neuron_L513-Coils-group-1.csv"));
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L513/Neuron_L513-Coils-group-2.csv"));
+        fileCoilList.append(QString("/usr/share/nymea/modbus/Neuron_L513/Neuron_L513-Coils-group-3.csv"));
```
mzanetti: same for all of those... 

instead of hardcoding the path, use QStandardPaths::standardLocation(QStandardPaths::GenericDataLocation).first() instead

unipi/unipi.pro
```diff
@@ -2,12 +2,32 @@ include(../plugins.pri)
 
 TARGET = $$qtLibraryTarget(nymea_devicepluginunipi)
 
+QT += \
+    network \
+    serialport \
+    core \
```
mzanetti: core isn't needed

unipi/neuron.cpp
Boernsman: Qt 

> Also, it doesn't build because of
> 
> Project ERROR: Unknown module(s) in QT: serialbus
> 
> propbably serialport is missing too
serialbus is not available for stretch, package needs to be delivered by ourself

unipi/neuron.cpp
Boernsman: changed it, but used .last() instead, first() is pointing to an user folder

Boernsman:
moved to a segregated repository

PR #114, opened 2019-05-26
New Plugin: WS2812FX

Boernsman:


ws2812fx/devicepluginws2812fx.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+            serialPort->setBaudRate(115200);
+            serialPort->setDataBits(QSerialPort::DataBits::Data8);
+            serialPort->setParity(QSerialPort::Parity::NoParity);
+            serialPort->setStopBits(QSerialPort::StopBits::OneStop);
+            serialPort->setFlowControl(QSerialPort::FlowControl::NoFlowControl);
+
+            if (!serialPort->open(QIODevice::ReadWrite)) {
+                qCWarning(dcWs2812fx()) << "Could not open serial port" << interface << serialPort->errorString();
+                return DeviceManager::DeviceSetupStatusFailure;
```
mzanetti: You're leaking the serialPort object here

ws2812fx/devicepluginws2812fx.cpp
```diff
@@ -0,0 +1,432 @@
...
+    Q_FOREACH(QSerialPortInfo port, QSerialPortInfo::availablePorts()) {
+
+        qCDebug(dcWs2812fx()) << "Found serial port:" << port.portName();
+        QString description = port.manufacturer() + " " + port.description();
+        DeviceDescriptor descriptor(deviceClassId, port.portName(), description);
+        ParamList parameters;
+        parameters.append(Param(ws2812fxDeviceSerialPortParamTypeId, port.portName()));
+        descriptor.setParams(parameters);
+        deviceDescriptors.append(descriptor);
+    }
```
mzanetti: Reconfigure isn't working with this. You want to set `deviceId` in the descriptor for already existing devices.

ws2812fx/devicepluginws2812fx.cpp
```diff
@@ -0,0 +1,432 @@
...
+        if (action.actionTypeId() == ws2812fxPowerActionTypeId) {
+            command.append("b ");
+            if (action.param(ws2812fxPowerActionPowerParamTypeId).value().toBool()) {
+                command.append("30");
+            } else {
+                command.append("0");
+            }
+            command.append("\r\n");
+            qDebug(dcWs2812fx()) << "Sending command" << command;
+            serialPort->write(command);
```
mzanetti: Hmm... you're never checking the return value of `write()`... I wonder what happens if you unplug the cable...

I'd recommend to create a function like
```
DeviceManager::DeviceError sendCommand(Device* device, const QByteArray &command)
{
    QSerialPort *serialPort = m_serialPorts.value(device);
    if (!serialPort)
        return DeviceManager::DeviceErrorDeviceNotFound;
    if (serialPort->write(command) != command.length()) {
        qCWarning(dcWs28812fx) << "Error writing to serial port";
        return DeviceManager::DeviceErrorHardwareFailure;
    }
    return DeviceManager::DeviceErrorAsync;
}
```

and then this whole if should become muuuch simpler.

ws2812fx/devicepluginws2812fx.cpp
```diff
@@ -0,0 +1,432 @@
...
+{
+    QSerialPort *serialPort =  static_cast<QSerialPort*>(sender());
+    Device *device = m_serialPorts.key(serialPort);
+
+    if (error != QSerialPort::NoError && serialPort->isOpen()) {
+        qCCritical(dcWs2812fx()) << "Serial port error:" << error << serialPort->errorString();
+        m_reconnectTimer->start();
+        serialPort->close();
+        device->setStateValue(ws2812fxConnectedStateTypeId, false);
+    }
```
mzanetti: Are you recovering from this somehow? Trying to re-open or so?

Boernsman: yes as soon as the reconnect timer triggers

mzanetti:
Jenkins, retest this please

ws2812fx/devicepluginws2812fx.cpp
```diff
@@ -0,0 +1,426 @@
...
+#include "devicepluginws2812fx.h"
+#include "plugininfo.h"
+
+DevicePluginWs2812fx ::DevicePluginWs2812fx ()
+{
+}
+
+Device::DeviceSetupStatus DevicePluginWs2812fx::setupDevice(Device *device)
+{
+    if(!m_reconnectTimer) {
```
mzanetti: It's probably better to just do this before returning Success... This way you have the timer running even if you fail later on.

PR #129, opened 2019-07-11
Add plug-in bose soundtouch

Boernsman:


bose/soundtouch.cpp
```diff
@@ -0,0 +1,781 @@
...
+    case KEY_VALUE_SHUFFLE_ON:
+        content.append("SHUFFLE_ON");
+        break;
+    case KEY_VALUE_SHUFFLE_OFF:
+        content.append("SHUFFLE_OFF");
+        break;
+    default:
+        qWarning(dcBose) << "key not yet implemented";
+        return;
+    }
```
mzanetti:
```
QXmlStreamWriter xml(&content);
xml.writeStartDocument("1.0");
xml.writeStartElement("key");
xml.writeAttribute("state", "Press");
xml.writeAttribute("sender", "Gabbo");
xml.writeCharacters("MUTE");
xml.writeEndElement();
xml.writeEndDocument();
```
In this particular case it actually turns out to be more to type than writing the xml hand-craftet, still, writing or parsing xml by hand is one of those things that will come back one day and kick your ass.

mzanetti: Oh I see you are using the stream writer for larger xml later on... Well then... fine with me for this size of snippet.

bose/soundtouch.cpp
```diff
@@ -0,0 +1,781 @@
...
+{
+    qDebug(dcBose) << "Bose websocket connected";
+    emit connectionChanged(true);
+}
+
+void SoundTouch::onWebsocketDisconnected()
+{
+    qDebug(dcBose) << "Bose websocket disconnected";
+    emit connectionChanged(false);
+    QTimer::singleShot(5000, [this](){
```
mzanetti: There is a "this" missing.... This will probably crash if you remove the SoundTouch object while this timer is running.


```
QTimer::singleShot(5000, this, [this](){
```
Using `this` as target object to connect to, will disconnect this signal/slot connection when `this` is deleted.. Think if it as connecting the timer to a lambda of this object, instead of connecting to a lambda without context, and hence without lifetime restriction.

mzanetti:
Jenkins, retest this please

PR #131, opened 2019-07-13
New plugin: generic interfaces

Boernsman:
Plug-in similar to generic elements but utilizing nymea interfaces


mzanetti:
Jenkins, retest this please

PR #138, opened 2019-08-07
Fix senic

Boernsman:


Boernsman:
Device tests have been successful

Boernsman:
added long press event and tested it successfully

PR #139, opened 2019-08-08
Tasmota: Add support for 3ch devices

myshoeshurt:


mzanetti:
Fails to  build with: PluginMetadata: DeviceClass "tasmotaSwitch" claims to implement interface "powerswitch" but doesn't implement event "pressed"

PR #144, opened 2019-08-27
New plug-in: systemmonitor

Boernsman:


nymea-plugins.pro
```diff
@@ -31,6 +31,13 @@ PLUGIN_DIRS = \
     osdomotics          \
     philipshue          \
     pushbullet          \
+    usbwde              \
+    multisensor         \
+    gpio                \
+    snapd               \
+    simulation          \
+    systemmonitor       \
+    keba                \
```
mzanetti: This change seems wrong. what happened?

systemmonitor/README.md
```diff
@@ -0,0 +1,3 @@
+# Systemmonitor
+
+
```
mzanetti: Can you add something here please? this will be used to generate webpage content. shouldn't be empty

PR #147, opened 2019-08-29
New Plug-In: Doorbird

Boernsman:


doorbird/deviceplugindoorbird.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+    qCWarning(dcDoorBird()) << "Cannot discover for deviceClassId" << info->deviceClassId();
+    info->finish(Device::DeviceErrorDeviceNotFound);
+}
+
+
+void DevicePluginDoorbird::startPairing(DevicePairingInfo *info)
+{
+    if (info->deviceClassId() == doorBirdDeviceClassId) {
+        qCDebug(dcDoorBird()) << "User and password. Login is \"user\" and \"password\".";
```
mzanetti: ???

mzanetti: Is that the default that can be changed in the device? If so, please change the text to be something like:

Please enter username and password for the DoorBird device. By default, this is \"username\" and \"password\" unless it has been changed on the DoorBird device.

Actually, I'd even prefer not having that in here at all and maybe add that information to the README instead...

doorbird/deviceplugindoorbird.cpp
```diff
@@ -0,0 +1,244 @@
...
+    if (info->deviceClassId() == doorBirdDeviceClassId) {
+        qCDebug(dcDoorBird()) << "confirm pairing called";
+
+        pluginStorage()->beginGroup(info->deviceId().toString());
+        pluginStorage()->setValue("username", username);
+        pluginStorage()->setValue("password", secret);
+        pluginStorage()->endGroup();
+
+        info->finish(Device::DeviceErrorNoError);
+        return;
```
mzanetti: ideally you'd try some call here to verify if it's correct...

doorbird/deviceplugindoorbird.cpp
```diff
@@ -0,0 +1,244 @@
...
+        pluginStorage()->endGroup();
+
+        qCDebug(dcDoorBird()) << "Device setup" << device->name() << username << password;
+        Doorbird *doorbird = new Doorbird(address, username, password, this);
+        connect(doorbird, &Doorbird::deviceConnected, this, &DevicePluginDoorbird::onDoorBirdConnected);
+        connect(doorbird, &Doorbird::eventReveiced, this, &DevicePluginDoorbird::onDoorBirdEvent);
+        connect(doorbird, &Doorbird::requestSent, this, &DevicePluginDoorbird::onDoorBirdRequestSent);
+        doorbird->connectToEventMonitor();
+        m_doorbirdConnections.insert(device, doorbird);
+        info->finish(Device::DeviceErrorNoError);
```
mzanetti: same here... if the user entered username/password wrong, he will be left with a seemingly functional device which just doesn't work... Not "the nymea experience" we're striving for...

Please do some authorized call here and set device error to Device::DeviceErrorAuthenticationFailure in case of failure with some informational message...

doorbird/doorbird.cpp
```diff
@@ -0,0 +1,460 @@
...
+#include <QUrl>
+#include <QUrlQuery>
+
+Doorbird::Doorbird(const QHostAddress &address, const QString &username, const QString &password, QObject *parent) :
+    QObject(parent),
+    m_address(address),
+    m_username(username),
+    m_password(password)
+{
+    m_networkAccessManager = new QNetworkAccessManager(this);
```
mzanetti: you should probably use hardwareManager's networkaccessmanager instead of creating a new one here... This isn't really critical at this point tho...

doorbird/doorbird.cpp
```diff
@@ -0,0 +1,460 @@
...
+    connect(m_networkAccessManager, &QNetworkAccessManager::authenticationRequired, this, [this](QNetworkReply *reply, QAuthenticator *authenticator) {
```
mzanetti: oh, I see, you're creating a new one because of this I guess...


Well, just note that you could also provide username and password in the authorize header directly, which would save the additional ping pong of Access denied etc...

doorbird/deviceplugindoorbird.cpp
```diff
@@ -0,0 +1,244 @@
...
+    }
+}
+
+
+void DevicePluginDoorbird::executeAction(DeviceActionInfo *info)
+{
+    Device *device = info->device();
+    Action action = info->action();
+
+    if (device->deviceClassId() == doorBirdDeviceClassId) {
```
mzanetti: error handling is quite bad in here too...

doorbird/deviceplugindoorbird.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+        break;
+    }
+}
+
+void DevicePluginDoorbird::onDoorBirdRequestSent(QUuid requestId, bool success)
+{
+    Doorbird *doorbird = static_cast<Doorbird *>(sender());
+
+    if (m_asyncActions.contains(requestId)) {
```
mzanetti: Doesn't look like executeAction would actually insert anything here...

doorbird/doorbird.pro
```diff
@@ -0,0 +1,13 @@
+include(../plugins.pri)
+
+QT += network
+
+TARGET = $$qtLibraryTarget(nymea_integrationplugindoorbird)
```
mzanetti: This line is not needed (but also does not harm)

doorbird/integrationplugindoorbird.cpp
```diff
@@ -0,0 +1,306 @@
...
+                    info->addThingDescriptor(ThingDescriptor);
+                }
+            }
+            serviceBrowser->deleteLater();
+            info->finish(Thing::ThingErrorNoError);
+        });
+        return;
+    } else {
+        qCWarning(dcDoorBird()) << "Cannot discover for ThingClassId" << info->thingClassId();
+        info->finish(Thing::ThingErrorThingNotFound);
```
mzanetti: Should probably be ThingErrorThingClassNotFound, but this will never be called anyways... Actually nowadays I don't even add the check for the thingClassId any more

PR #149, opened 2019-09-04
New plugin: Sonos

Boernsman:


sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,526 @@
...
+            pluginStorage()->beginGroup(info.deviceId().toString());
+            pluginStorage()->setValue("access_token", accessToken);
+            pluginStorage()->setValue("refresh_token", refreshToken);
+            pluginStorage()->endGroup();
+
+            if (jsonDoc.toVariant().toMap().contains("expires_in")) {
+                int expireTime = jsonDoc.toVariant().toMap().value("expires_in").toInt();
+                qCDebug(dcSonos()) << "expires at" << QDateTime::currentDateTime().addSecs(expireTime).toString();
+                //m_tokenRefreshTimer->start((expireTime - 20) * 1000);
+                //TODO
```
mzanetti: TBD before landing :)

mzanetti:
@rhythm-section Please test this one.

sonos/devicepluginsonos.cpp
```diff
@@ -331,6 +332,55 @@ void DevicePluginSonos::executeAction(DeviceActionInfo *info)
...
+void DevicePluginSonos::executeBrowserItem(BrowserActionInfo *info)
+{
+    Device *parentDevice = myDevices().findById(info->device()->parentId());
+    Sonos *sonosConnection = m_sonosConnections.value(parentDevice);
+    if (!sonosConnection)
+        return;
+
+    QString groupId = info->device()->paramValue(sonosGroupDeviceGroupIdParamTypeId).toString();
+    QUuid requestId = sonosConnection->loadFavorite(groupId, info->browserAction().itemId());
+    m_pendingBrowserExecution.insert(requestId, info);
```
mzanetti: If you're storing the info pointer locally you should either connect to it's aborted() signal and remove it from the list, or, you store it inside a QPointer<BrowserItem> and check that for isNull() before accessing it...

You must not access a info object after it has finished/aborted

sonos/devicepluginsonos.cpp
```diff
@@ -366,25 +416,45 @@ void DevicePluginSonos::onHouseholdIdsReceived(QList<QString> householdIds)
...
-    Q_UNUSED(householdId);
-    foreach(Sonos::FavouriteObject favourite, favourites)  {
-        qDebug(dcSonos()) << "Favourite: " << favourite.name << favourite.description;
+    if (m_pendingBrowseResult.contains(householdId)) {
+        BrowseResult *result = m_pendingBrowseResult.take(householdId);
+        if (!result)
+            return;
+
+        foreach(Sonos::FavouriteObject favourite, favourites)  {
+            BrowserItem item;
```
mzanetti: Using MediaBrowserItem instead of just BrowserItem in this line might be enough already..  You might set addtional icons in that case.

sonos/devicepluginsonos.h
```diff
@@ -49,18 +49,26 @@ class DevicePluginSonos : public DevicePlugin
...
-     QHash<DeviceId, Sonos *> m_setupSonosConnections;
+    QHash<DeviceId, Sonos *> m_setupSonosConnections;
     QHash<Device *, Sonos *> m_sonosConnections;
     QList<QByteArray> m_householdIds;
 
     QByteArray m_sonosConnectionAccessToken;
     QByteArray m_sonosConnectionRefreshToken;
 
     QHash<QUuid, QPointer<DeviceActionInfo> > m_pendingActions;
+    QHash<QString, BrowseResult *> m_pendingBrowseResult;
```
mzanetti: Same here ofc... Either a `QHash<QString QPointer<BrowseResult> >` or clean up yourself by connecting to `aborted()`

sonos/devicepluginsonos.json
```diff
@@ -160,49 +167,56 @@
...
                             "id": "85d7126a-b123-4a28-aeb4-d84bcfb4d14f",
                             "name": "skipNext",
-                            "displayName": "skipNext"
+                            "displayName": "Skip Next"
+                        }
+                    ],
+                    "browserItemActionTypes": [
+                        {
+                            "id": "b056af0f-4d5b-4f75-b0dd-e73246d3a83f",
+                            "name": "updateLibrary",
```
mzanetti: I don't see this handled in the plugin...

sonos/devicepluginsonos.cpp
```diff
@@ -331,6 +332,55 @@ void DevicePluginSonos::executeAction(DeviceActionInfo *info)
...
+        return;
+
+    QString groupId = info->device()->paramValue(sonosGroupDeviceGroupIdParamTypeId).toString();
+    QUuid requestId = sonosConnection->loadFavorite(groupId, info->browserAction().itemId());
+    m_pendingBrowserExecution.insert(requestId, info);
+}
+
+void DevicePluginSonos::executeBrowserItemAction(BrowserItemActionInfo *info)
+{
+    Q_UNUSED(info)
```
mzanetti: Shouldn't be empty if you're defining BrowserItemActionTypes...

sonos/devicepluginsonos.json
mzanetti: Actually, it's not set to any item either afaict.

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,599 @@
...
+        m_pendingBrowseResult.insert(householdId, result);
+    } else {
+        //TODO add media browsing
+        result->finish(Device::DeviceErrorItemNotFound);
+    }
+}
+
+void DevicePluginSonos::browserItem(BrowserItemResult *result)
+{
+    Q_UNUSED(result)
```
mzanetti: Oh. also, this should not be empty... Needs to return a single item with the given id or it won't work in rules (Well actually it would work, but the client app has no chance in fetching details for this item if the user wants to edit a rule)

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,639 @@
...
+    }
+    info->finish(Device::DeviceErrorDeviceClassNotFound);
+}
+
+void DevicePluginSonos::browseDevice(BrowseResult *result)
+{
+    Device *parentDevice = myDevices().findById(result->device()->parentId());
+    Sonos *sonosConnection = m_sonosConnections.value(parentDevice);
+    if (!sonosConnection)
+        return;
```
mzanetti: You need to `finish()` the result or it'll cause the app to wait for 30 secs for the timeout.

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,639 @@
...
+        item.setId("/favorites");
+        item.setIcon(BrowserItem::BrowserIconFavorites);
+        item.setExecutable(false);
+        item.setBrowsable(true);
+        item.setDisplayName("Favorites");
+        result->addItem(item);
+        result->finish(Device::DeviceErrorNoError);
+    } else if (result->itemId() == "/favorites") {
+        sonosConnection->getFavorites(householdId);
+        m_pendingBrowseResult.insert(householdId, result);
```
mzanetti: hmm... This seems... I don't think you can use `hoseholdId` to identify calls... this will clash if 2 clients browse simultaneously because both will use the same householdId and overwrite each other's results... I'm afraid you need to change the api for Sonos::getFavorites() etc to track calls by a unique id... I'd probably use an integer generated inside getFavorites.. like this:

```
int Sonos::getFavorites() {
  int id = m_currentId++;
  QNetworkReply *reply = networkManager()->get(...);
  connect(reply, &finished, this, [reply, id]() {
    ...
    emit favoritesReceived(id, result);
  }
}
```

and in here

```
  int id = sonosConnection->getFavorites(householdId);
  m_pendingBrowseResult.insert(id, result);
  connect(result, &timeout, this, [this, id](){ m_pendingBrowseResults.remove(id); });
```

and 

```
void DevicePluginSonos::favoritesRecived(int id, Favorites favorites) {
   BrowseResult *result = m_pendingBrowseResults.value(id);
   ...
   result->finish(...);
}
```

Also make sure to always finish() browse calls! I see quite some paths here that leave it time out...

And, you're mixing en_US and en_GB in getFavorites and favouritesReceived ;)

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,639 @@
...
+        result->finish(Device::DeviceErrorItemNotFound);
+    }
+}
+
+void DevicePluginSonos::browserItem(BrowserItemResult *result)
+{
+    Device *parentDevice = myDevices().findById(result->device()->parentId());
+    Sonos *sonosConnection = m_sonosConnections.value(parentDevice);
+    if (!sonosConnection)
+        return;
```
mzanetti: forgot to finish() the result

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,639 @@
...
+                item.setExecutable(true);
+                item.setBrowsable(false);
+                if (!favourite.imageUrl.isEmpty()) {
+                    item.setThumbnail(favourite.imageUrl);
+                } else {
+                    item.setIcon(BrowserItem::BrowserIconFavorites);
+                }
+                item.setDisplayName(favourite.name);
+                item.setDescription(favourite.description);
+                result->finish(item);
```
mzanetti: you can `return` here to spare looping over the rest of the items needlessly...

sonos/devicepluginsonos.cpp
```diff
@@ -0,0 +1,639 @@
...
+    if (!sonosConnection)
+        return;
+
+    qCDebug(dcSonos()) << "Browser Item" << result->itemId();
+    QString householdId = result->device()->paramValue(sonosGroupDeviceHouseholdIdParamTypeId).toString();
+    if (result->itemId().startsWith("/favorites")) {
+        sonosConnection->getFavorites(householdId);
+        m_pendingBrowserItemResult.insert(householdId, result);
+        connect(result, &BrowserItemResult::aborted,[householdId, this](){m_pendingBrowserItemResult.remove(householdId);});
+    }
```
mzanetti: Add the `// TODO: Media browsing here too`

sonos/devicepluginsonos.cpp
mzanetti: Actually now I see that non-browsing related calls use a QUuid as requestId, that works as well...

PR #183, opened 2019-11-10
Http commander: Add http server

Boernsman:

* Make sure the pull request's title is of format "Plugin name: Add support for xyz" or "New plugin: Plugin name"

* Did you test the changes on hardware, if not (e.g. absence of required hardware), please mention a person to confirm it has been tested.

* Did you update the plugin's README.md accordingly?

* Did you update translations (`cd builddir && make lupdate`)?

* If you added a new plugin, should it be added to nymea-plugins-all or nymea-plugins-maker?


httpcommander/devicepluginhttpcommander.cpp
```diff
@@ -69,6 +69,24 @@ void DevicePluginHttpCommander::setupDevice(DeviceSetupInfo *info)
...
+
+        QString interfaces;
+        foreach (const QNetworkInterface &interface,  QNetworkInterface::allInterfaces()) {
+            foreach (QNetworkAddressEntry entry, interface.addressEntries()) {
+                if (entry.ip().protocol() == QAbstractSocket::IPv4Protocol && entry.ip().toString() != "127.0.0.1") {
+                    interfaces.append(entry.ip().toString());
+                }
+            }
+        }
+        device->setStateValue(httpServerServerAddressStateTypeId, interfaces);
```
mzanetti: Hmm... you're appending multiple addresses here? but no separator... 
This is probably meant to be informational to the user what IP addresses it has...

So, there's a type StringList which would take a QStringList

Also it seems a bit shoehorned in... That information should probably be in the system settings -> networking instead, but it's not in there yet so I guess this is fine for now...

Also, would it make sense to instead have a param of with 
possibleValues: ["Localhost", "Any"]. Not saying you have to add that now, just thinking out loud as we probably want to convert the http server to a system feature (hardware resource) eventually.

httpcommander/httpsimpleserver.cpp
```diff
@@ -0,0 +1,104 @@
...
+#include <QDebug>
+#include <QDateTime>
+#include <QUrlQuery>
+#include <QRegExp>
+#include <QStringList>
+
+HttpSimpleServer::HttpSimpleServer(QObject *parent):
+    QTcpServer(parent)
+{
+    listen(QHostAddress::Any, 7777);
```
mzanetti: erm, isn't there a param for the port? Seems like that one isn't used at all and this will fail for more than one servers

httpcommander/devicepluginhttpcommander.cpp
```diff
@@ -60,175 +47,82 @@ void DevicePluginHttpCommander::setupDevice(DeviceSetupInfo *info)
...
+            if (method == "GET") {
+                reply = hardwareManager()->networkManager()->get(QNetworkRequest(url));
+            } else if (method == "POST") {
+                reply = hardwareManager()->networkManager()->post(QNetworkRequest(url), payload);
+            } else if (method == "PUT") {
+                reply = hardwareManager()->networkManager()->put(QNetworkRequest(url), payload);
+            } else if (method == "DELETE") {
+                reply = hardwareManager()->networkManager()->deleteResource(QNetworkRequest(url));
+            }
+            connect(reply, &QNetworkReply::finished, this, [device, reply, this](){
```
mzanetti: should be:

```
connect(reply, &QNetworkReply::finished, reply, &QNetworkReply::deleteLater());
connect(reply &QNetworkReply::finished, device, [device, reply, this]() {
...
}
```

Reasoning: The way you have it currently will crash if a network operation is started and the device is removed from the system before the network reply returns because the captured `device` is deleted.

The solution to that is that you connect the lambda to `device` instead of `this`. That way it won't be executed any more if the device is deleted. There's one new issue in that case however, the `reply->deleteLater()` won't be called any more and you'd leak the device. Because of that you connect the `reply->finished()` signal to `the reply->deleteLater()` slot, so that line will be executed in any case when the reply finished, regardless if the device has been deleted by then, or even if the plugin has been unloaded.

httpcommander/devicepluginhttpcommander.cpp
```diff
@@ -60,175 +47,82 @@ void DevicePluginHttpCommander::setupDevice(DeviceSetupInfo *info)
...
+                int status = reply->attribute(QNetworkRequest::HttpStatusCodeAttribute).toInt();
+
+                device->setStateValue(httpRequestResponseStateTypeId, data);
+                device->setStateValue(httpRequestStatusStateTypeId, status);
+
+                // Check HTTP status code
+                if (status != 200 || reply->error() != QNetworkReply::NoError) {
+                    qCWarning(dcHttpCommander()) << "Request error:" << status << reply->errorString();
+                }
+                reply->deleteLater();
```
mzanetti: you can drop this line if you do the above.

PR #207, opened 2019-12-30
New Plug-In: OpenUV

Boernsman:

* Make sure the pull request's title is of format "Plugin name: Add support for xyz" or "New plugin: Plugin name"

* Did you test the changes on hardware, if not (e.g. absence of required hardware), please mention a person to confirm it has been tested.

* Did you update the plugin's README.md accordingly?

* Did you update translations (`cd builddir && make lupdate`)?

* If you added a new plugin, should it be added to nymea-plugins-all or nymea-plugins-maker?

### DIFF
--- a/openuv/devicepluginopenuv.cpp
+++ b/openuv/devicepluginopenuv.cpp
@@ -150,9 +150,9 @@ void DevicePluginOpenUv::getUvIndex(Device *device)
         device->setStateValue(openUvUvMaxStateTypeId, result["uv_max"].toDouble());
         device->setStateValue(openUvOzoneStateTypeId, result["ozone"].toDouble());
 
-        device->setStateValue(openUvUvTimeStateTypeId, QDateTime::fromString(result["uv_time"].toString(),Qt::DateFormat::ISODateWithMs).toSecsSinceEpoch());
-        device->setStateValue(openUvOzoneTimeStateTypeId, QDateTime::fromString(result["ozone_time"].toString(),Qt::DateFormat::ISODateWithMs).toSecsSinceEpoch());
-        device->setStateValue(openUvUvMaxTimeStateTypeId, QDateTime::fromString(result["uv_max_time"].toString(),Qt::DateFormat::ISODateWithMs).toSecsSinceEpoch());
+        device->setStateValue(openUvUvTimeStateTypeId, QDateTime::fromString(result["uv_time"].toString(),Qt::DateFormat::ISODate).toMSecsSinceEpoch() / 1000);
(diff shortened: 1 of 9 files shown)


openuv/devicepluginopenuv.cpp
```diff
@@ -0,0 +1,173 @@
...
+        descriptor.setParams(params);
+        info->addDeviceDescriptor(descriptor);
+        info->finish(Device::DeviceErrorNoError);
+    });
+}
+
+void DevicePluginOpenUv::setupDevice(DeviceSetupInfo *info)
+{
+    if (m_apiKey.isEmpty() || m_apiKey == "-") {
+        info->finish(Device::DeviceErrorSetupFailed, tr("API key is missing, add your OpenUV API key in the plug-in configs"));
```
mzanetti: needs to be `QT_TR_NOOP(...)` 

`tr()` is auto-translated by Qt but that doesn't work for us as we need to translate it manually on the fly depending on the connected client.

openuv/devicepluginopenuv.cpp
```diff
@@ -0,0 +1,173 @@
...
+    QUrlQuery query;
+    query.addQueryItem("lat", device->paramValue(openUvDeviceLatitudeParamTypeId).toString());
+    query.addQueryItem("lng", device->paramValue(openUvDeviceLongitudeParamTypeId).toString());
+    url.setQuery(query);
+    QNetworkRequest request;
+    request.setUrl(url);
+    request.setRawHeader("x-access-token", m_apiKey);
+
+    QNetworkReply *reply = hardwareManager()->networkManager()->get(request);
+    connect(reply, &QNetworkReply::finished, this, [reply, device, this] {
```
mzanetti: you're capturing `device`, use it as receiver object instead of `this` or it'll crash if the user removes the device while a get call is pending.

openuv/devicepluginopenuv.cpp
```diff
@@ -0,0 +1,173 @@
...
+        reply->deleteLater();
```
mzanetti: ... which implies the delete should be connected individually to happen in any case.

openuv/devicepluginopenuv.json
```diff
@@ -0,0 +1,162 @@
+{
+    "name": "OpenUv",
+    "displayName": "OpenUV",
+    "id": "9b7d9cc8-77df-4197-a6fc-8a365747a3b1",
+    "paramTypes": [
+        {
+            "id": "cbb8e9d1-22b6-40ab-bbff-26abb2ed416f",
+            "name": "apiKey",
```
mzanetti: Same discussion as in the other plugin applies... Shouldn't this be a pairing option of enterPin instead?

mzanetti:
Also fails to build on older qt versions

PR #232, opened 2020-02-14
Bose: Add support for presets

Boernsman:

### DIFF
--- a/bose/devicepluginbose.cpp
+++ b/bose/devicepluginbose.cpp
@@ -87,6 +87,7 @@ void DevicePluginBose::setupDevice(DeviceSetupInfo *info)
         connect(soundTouch, &SoundTouch::bassCapabilitiesReceived, this, &DevicePluginBose::onBassCapabilitiesObjectReceived);
         connect(soundTouch, &SoundTouch::zoneReceived, this, &DevicePluginBose::onZoneObjectReceived);
         connect(soundTouch, &SoundTouch::requestExecuted, this, &DevicePluginBose::onRequestExecuted);
+        connect(soundTouch, &SoundTouch::presetsReceived, this, &DevicePluginBose::onPresetsReceived);
         m_soundTouch.insert(info->device(), soundTouch);
         return info->finish(Device::DeviceErrorNoError);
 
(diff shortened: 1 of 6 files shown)


bose/devicepluginbose.cpp
```diff
@@ -490,3 +556,27 @@ void DevicePluginBose::onZoneObjectReceived(QUuid requestId, ZoneObject zone)
         qDebug(dcBose()) << "-> member:" << member.deviceID;
     }
 }
+
+void DevicePluginBose::onPresetsReceived(QUuid requestId, QList<PresetObject> presets)
```
mzanetti: const QUuid &requestId, const QList<PresetObject> &presets

bose/devicepluginbose.h
```diff
@@ -87,6 +87,7 @@ private slots:
    void onBassCapabilitiesObjectReceived(QUuid requestId, BassCapabilitiesObject bassCapabilities);
    void onGroupObjectReceived(QUuid requestId, GroupObject group);
    void onZoneObjectReceived(QUuid requestId, ZoneObject zone);
+   void onPresetsReceived(QUuid requestId, QList<PresetObject> presets);
```
mzanetti: hmm, you should use `const &` for all of those arguments in all those slots... otherwise it will copy all those things on every call... Not insanely huge, but still a performance and memory penalty we can easily avoid

bose/soundtouch.cpp
```diff
@@ -174,7 +174,12 @@ QUuid SoundTouch::setKey(KEY_VALUE keyValue)
     QXmlStreamWriter xml(&content);
     xml.writeStartDocument("1.0");
     xml.writeStartElement("key");
-    xml.writeAttribute("state", "press");
+    if (pressed) {
+        xml.writeAttribute("state", "press");
+    } else {
+        xml.writeAttribute("state", "release");
+    }
+
```
mzanetti: Wondering... I only see you calling this with "true", is actively releasing not required?

bose/soundtouch.h
```diff
@@ -95,7 +95,7 @@ class SoundTouch : public QObject
     void zoneReceived(QUuid requestId, ZoneObject);
     void bassCapabilitiesReceived(QUuid requestId, BassCapabilitiesObject bassCapabilities);
     void bassReceived(QUuid requestId, BassObject bass);
-    void presetsReceived(QUuid requestId, PresetObject presets);
+    void presetsReceived(QUuid requestId, QList<PresetObject> presets);
```
mzanetti: same here... please use `const &` for the parameters

mzanetti:
Also, the pull request title doesn't follow the checklist format ;)

PR #256, opened 2020-04-14
OneWire: Fix manual discovery of connected devices

mzanetti:


onewire/integrationpluginonewire.cpp
```diff
@@ -59,15 +59,16 @@ void IntegrationPluginOneWire::discoverThings(ThingDiscoveryInfo *info)
                 //devices cannot be discovered since auto mode is enabled
                 return info->finish(Thing::ThingErrorNoError);
```
Boernsman, line 60: ah that is the issue, I think the user should get feedback here why there has been no device discovered. and the discovery be finished with an error. Also multiple interfaces is not supported because it finishes already with the first one

PR #259, opened 2020-04-16
New Plugin: Home Connect

Boernsman:


mzanetti:
Fails to load:

 W | ThingManager: Could not load plugin data of "/usr/lib/arm-linux-gnueabihf/nymea/plugins/libnymea_integrationpluginhomeconnect.so" 
 "Cannot load library /usr/lib/arm-linux-gnueabihf/nymea/plugins/libnymea_integrationpluginhomeconnect.so: (/usr/lib/arm-linux-gnueabihf/nymea/plugins/libnymea_integrationpluginhomeconnect.so: undefined symbol: _ZN28IntegrationPluginHomeConnect23onReceivedActiveProgramERK7QStringS2_RK5QHashIS0_8QVariantE)"
 W | ThingManager: Error loading plugin: "/usr/lib/arm-linux-gnueabihf/nymea/plugins/libnymea_integrationpluginhomeconnect.so"

homeconnect/homeconnect.cpp
```diff
@@ -0,0 +1,820 @@
...
+
+    QUrl url(m_baseAuthorizationUrl);
+    QUrlQuery queryParams;
+    queryParams.addQueryItem("client_id", m_clientKey);
+    queryParams.addQueryItem("redirect_uri", m_redirectUri);
+    queryParams.addQueryItem("response_type", "code");
+    queryParams.addQueryItem("scope", scope);
+    queryParams.addQueryItem("state", QUuid::createUuid().toString());
+    queryParams.addQueryItem("nonce", QUuid::createUuid().toString());
+    m_codeChallenge = QUuid::createUuid().toString().remove('{').remove('}').remove("-");
```
mzanetti, line 96: .remove(QRegExp("[{}-]")

homeconnect/homeconnect.cpp
```diff
@@ -0,0 +1,820 @@
...
+
+void HomeConnect::onRefreshTimeout()
+{
+    qCDebug(dcHomeConnect) << "Refresh authentication token";
+    getAccessTokenFromRefreshToken(m_refreshToken);
+}
+
+bool HomeConnect::checkStatusCode(int status, const QByteArray &payload)
+{
+    //TODO emit (dis)connected, (un)authenticated,
```
mzanetti, line 112: hmm... seems a somewhat important todo... Much to be done there or could we have it fixed still?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+        bool simulationMode =  configValue(homeConnectPluginSimulationModeParamTypeId).toBool();
+        bool controlEnabled =  configValue(homeConnectPluginControlEnabledParamTypeId).toBool();
+        QByteArray clientKey = apiKeyStorage()->requestKey("homeconnect").data("clientKey");
+        QByteArray clientSecret = apiKeyStorage()->requestKey("homeconnect").data("clientSecret");
+        HomeConnect *homeConnect = new HomeConnect(hardwareManager()->networkManager(), clientKey, clientSecret, simulationMode, this);
+        QString scope = "IdentifyAppliance Monitor Settings Dishwasher Washer Dryer WasherDryer Refrigerator Freezer WineCooler CoffeeMaker Hood CookProcessor";
+        if (controlEnabled)
+            scope.append(" Control");
+        QUrl url = homeConnect->getLoginUrl(QUrl("https://127.0.0.1:8888"), scope);
+        qCDebug(dcHomeConnect()) << "HomeConnect url:" << url;
```
mzanetti, line 154: is this printing the api key to the log?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+        info->finish(Thing::ThingErrorCreationMethodNotSupported);
+    }
+}
+
+void IntegrationPluginHomeConnect::confirmPairing(ThingPairingInfo *info, const QString &username, const QString &secret)
+{
+    Q_UNUSED(username);
+
+    if (info->thingClassId() == homeConnectConnectionThingClassId) {
+        qCDebug(dcHomeConnect()) << "Redirect url is" << secret;
```
mzanetti, line 169: same... please avoid printing sensitive data to the log. Ideally you'd obfuscate sensitive information but still keep a debug print with enough data to be useful for debugging. I know, it's a bit of extra efford... But it pays back when you don't have to play log-ping-pong with users.

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+
+        HomeConnect *homeConnect = m_setupHomeConnectConnections.value(info->thingId());
+        if (!homeConnect) {
+            qWarning(dcHomeConnect()) << "No HomeConnect connection found for device:"  << info->thingName();
+            m_setupHomeConnectConnections.remove(info->thingId());
+            homeConnect->deleteLater();
+            info->finish(Thing::ThingErrorHardwareFailure);
+            return;
+        }
+        qCDebug(dcHomeConnect()) << "Authorization code" << authorizationCode;
```
mzanetti, line 182: :)

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+
+void IntegrationPluginHomeConnect::startPairing(ThingPairingInfo *info)
+{
+    if (info->thingClassId() == homeConnectConnectionThingClassId) {
+
+        bool simulationMode =  configValue(homeConnectPluginSimulationModeParamTypeId).toBool();
+        bool controlEnabled =  configValue(homeConnectPluginControlEnabledParamTypeId).toBool();
+        QByteArray clientKey = apiKeyStorage()->requestKey("homeconnect").data("clientKey");
+        QByteArray clientSecret = apiKeyStorage()->requestKey("homeconnect").data("clientSecret");
+        HomeConnect *homeConnect = new HomeConnect(hardwareManager()->networkManager(), clientKey, clientSecret, simulationMode, this);
```
mzanetti, line 149: reconfigure case not respected... This will leak on reconfigure...

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+        QString scope = "IdentifyAppliance Monitor Settings Dishwasher Washer Dryer WasherDryer Refrigerator Freezer WineCooler CoffeeMaker Hood CookProcessor";
+        if (controlEnabled)
+            scope.append(" Control");
+        QUrl url = homeConnect->getLoginUrl(QUrl("https://127.0.0.1:8888"), scope);
+        qCDebug(dcHomeConnect()) << "HomeConnect url:" << url;
+        info->setOAuthUrl(url);
+        info->finish(Thing::ThingErrorNoError);
+        m_setupHomeConnectConnections.insert(info->thingId(), homeConnect);
```
mzanetti, line 157: to be precise, this is the line that leaks on reconnect as you're replacing the old one in the map without deleting it and thus losing all references to it.

Still I'd fix it above where I'd do something like

```
HomeConnect *homeConnect = m_setupHomeConnections.value(info->thingId());
if (!homeConnect) {
  homeConnect *homeConnect = new HomeConnect(...);
  m_setupHomeConnections.insert(info->thingId(), homeConnect);
}
```

and just remove the insert here...

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+        if (!homeConnect) {
+            qWarning(dcHomeConnect()) << "No HomeConnect connection found for device:"  << info->thingName();
+            m_setupHomeConnectConnections.remove(info->thingId());
+            homeConnect->deleteLater();
+            info->finish(Thing::ThingErrorHardwareFailure);
+            return;
+        }
+        qCDebug(dcHomeConnect()) << "Authorization code" << authorizationCode;
+        homeConnect->getAccessTokenFromAuthorizationCode(authorizationCode);
+        connect(homeConnect, &HomeConnect::authenticationStatusChanged, this, [info, this](bool authenticated){
```
mzanetti, line 184: I think this might mess stuff up on reconfigure as an existing homeConnect object would keep being connected to this lambda until destroyed. So a reconfigure would connect it a second time.

One easy fix would be to use `info` as the third argument instead of `this`, then this connection would vanish when the setup finishes and a reconfigure can just reconnect it again when needed.

homeconnect/integrationpluginhomeconnect.h
```diff
@@ -0,0 +1,108 @@
...
+    void postSetupThing(Thing *thing) override;
+    void executeAction(ThingActionInfo *info) override;
+    void thingRemoved(Thing *thing) override;
+
+    void browseThing(BrowseResult *result) override;
+    void browserItem(BrowserItemResult *result) override;
+    void executeBrowserItem(BrowserActionInfo *info) override;
+
+private:
+    PluginTimer *m_pluginTimer5sec = nullptr;
```
mzanetti, line 63: seems unused, isn't it?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+        });
+    }
+
+    if (thing->thingClassId() == homeConnectConnectionThingClassId) {
+        HomeConnect *homeConnect = m_homeConnectConnections.value(thing);
+        homeConnect->getHomeAppliances();
+        homeConnect->connectEventStream();
+        thing->setStateValue(homeConnectConnectionConnectedStateTypeId, true);
+        thing->setStateValue(homeConnectConnectionLoggedInStateTypeId, true);
+        //TBD Set user name
```
mzanetti, line 295: I've checked the home connect api and there doesn't seem to be any user info either...

Strange thing is that virtually in all apis I've worked with so far, such an info was available. However, given that it apparently isn't I'm tempted to make the userDisplayName state in the account interface optional. WDYT?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+    } else if ((thing->thingClassId() == dryerThingClassId) ||
+               (thing->thingClassId() == fridgeThingClassId) ||
+               (thing->thingClassId() == washerThingClassId) ||
+               (thing->thingClassId() == dishwasherThingClassId) ||
+               (thing->thingClassId() == coffeeMakerThingClassId) ||
+               (thing->thingClassId() == ovenThingClassId) ||
+               (thing->thingClassId() == hoodThingClassId) ||
+               (thing->thingClassId() == cleaningRobotThingClassId) ||
+               (thing->thingClassId() == cookTopThingClassId)) {
```
mzanetti, line 304: couldn't you just use `if (m_idParamTypesMap.contains(info->thing()->thingClassId())` instead?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+
+    HomeConnect *homeConnect = m_homeConnectConnections.value(myThings().findById(thing->parentId()));
+    if (!homeConnect)
+        return;
+    QString haid = thing->paramValue(m_idParamTypeIds.value(thing->thingClassId())).toString();
+    homeConnect->getProgramsAvailable(haid);
+    connect(homeConnect, &HomeConnect::receivedAvailablePrograms, result, [result, this] (const QString &haid, const QStringList &programs) {
+        if (result->thing()->paramValue(m_idParamTypeIds.value(result->thing()->thingClassId())).toString() == haid) {
+            if (programs.contains(result->item().id()))
+                result->finish(Thing::ThingErrorNoError);
```
mzanetti, line 535: I don't see you actually adding the result here, are you? wouldn't this even crash?

result->item() is a nullptr until you set it. And if you return NoError you should really set it.

Try to create a rule where you run a browser entry. Then edit this rule and see if stuff works.

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,978 @@
...
+    } else if (key.contains("Refrigeration.Common.Setting.SabbathMode")) {
+    } else if (key.contains("Refrigeration.Common.Setting.VacationMode")) {
+    } else if (key.contains("Refrigeration.Common.Setting.FreshMode")) {
+    } else if (key.contains("Cooking.Common.Setting.Lighting")) {
+    } else if (key.contains("Cooking.Common.Setting.LightingBrightness")) {
+    } else if (key.contains("BSH.Common.Setting.AmbientLightEnabled")) {
+    } else if (key.contains("BSH.Common.Setting.AmbientLightBrightness")) {
+    } else if (key.contains("BSH.Common.Setting.AmbientLightColor")) {
+    } else if (key.contains("BSH.Common.Setting.AmbientLightCustomColor")) {
+    }
```
mzanetti, line 765: Ist das Kunst oder kann das weg?

mzanetti (changes requested):
Pretty nice plugin. Needs some fixes tho. And a couple of questions.

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+            info->finish(Thing::ThingErrorAuthenticationFailure, tr("Client key and/or seceret is not available."));
+            return;
+        } else {
+            qCDebug(dcHomeConnect()) << "Using API client secret and key from API key provider";
+        }
+        HomeConnect *homeConnect = new HomeConnect(hardwareManager()->networkManager(), clientKey, clientSecret, simulationMode, this);
+        QString scope = "IdentifyAppliance Monitor Settings Dishwasher Washer Dryer WasherDryer Refrigerator Freezer WineCooler CoffeeMaker Hood CookProcessor";
+        if (controlEnabled) {
+            scope.append(" Control");
+            qCDebug(dcHomeConnect()) << "Conrol scope is enabled";
```
mzanetti, line 171: typo control

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+                m_setupHomeConnectConnections.insert(info->thingId(), homeConnect);
+                connect(info, &ThingPairingInfo::aborted, this, [info, this] {m_setupHomeConnectConnections.take(info->thingId())->deleteLater();});
+                info->finish(Thing::ThingErrorNoError);
+            } else {
+                qCDebug(dcHomeConnect()) << "Got online check error" << reply->errorString();
+                info->finish(Thing::ThingErrorSetupFailed, tr("HomeConnect server not reachable, please check the internet connection"));
+            }
+        });
+    } else {
+        qCWarning(dcHomeConnect()) << "Unhandled pairing metod!";
```
mzanetti, line 192: typo method

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+{
+    Q_UNUSED(username);
+
+    if (info->thingClassId() == homeConnectAccountThingClassId) {
+        QUrl url(secret);
+        QUrlQuery query(url);
+        QByteArray authorizationCode = query.queryItemValue("code").toLocal8Bit();
+        if (authorizationCode.isEmpty()) {
+            qCWarning(dcHomeConnect()) << "No authorization code received.";
+            info->finish(Thing::ThingErrorSetupFailed);
```
mzanetti, line 207: missing return here?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+
+void IntegrationPluginHomeConnect::setupThing(ThingSetupInfo *info)
+{
+    Thing *thing = info->thing();
+
+    if (thing->thingClassId() == homeConnectAccountThingClassId) {
+        bool simulationMode =  configValue(homeConnectPluginSimulationModeParamTypeId).toBool();
+        HomeConnect *homeConnect;
+        if (m_setupHomeConnectConnections.keys().contains(thing->id())) {
```
mzanetti, line 241: will this behave properly on reconfiguration? Wouldn't you need to do the stuff below in that case?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+            }
+            if (clientKey.isEmpty() || clientSecret.isEmpty()) {
+                info->finish(Thing::ThingErrorAuthenticationFailure, tr("Client key and/or seceret is not available."));
+                return;
+            }
+            homeConnect = new HomeConnect(hardwareManager()->networkManager(), clientKey, clientSecret, simulationMode,  this);
+            homeConnect->getAccessTokenFromRefreshToken(refreshToken);
+            m_asyncSetup.insert(homeConnect, info);
+        }
+        connect(homeConnect, &HomeConnect::connectionChanged, this, &IntegrationPluginHomeConnect::onConnectionChanged);
```
mzanetti, line 272: this on the other hand should probably not be done in the reconfigure case, should it?

homeconnect/integrationpluginhomeconnect.cpp
```diff
@@ -0,0 +1,1041 @@
...
+
+    HomeConnect *homeConnect = m_homeConnectConnections.value(myThings().findById(thing->parentId()));
+    if (!homeConnect)
+        return;
+    QString haid = thing->paramValue(m_idParamTypeIds.value(thing->thingClassId())).toString();
+    homeConnect->getProgramsAvailable(haid);
+    connect(homeConnect, &HomeConnect::receivedAvailablePrograms, result, [result, this] (const QString &haid, const QStringList &programs) {
+        if (result->thing()->paramValue(m_idParamTypeIds.value(result->thing()->thingClassId())).toString() == haid) {
+            if (programs.contains(result->item().id())) {
+                result->item().setDisplayName(result->item().id());
```
mzanetti, line 558: item() returns a copy of the BrowserItem... calling setDisplayName() will vanish into nirvana.
You do need to create a BrowserItem and call info->finish(item) with it.

mzanetti (changes requested):

homeconnect/integrationpluginhomeconnect.cpp
Boernsman: Since a reconfigure is now envoking the removeThing method there should be no difference to a ordinary setup

mzanetti: erm, no, we changed the reconfigure to *not* invoke removeThing any more

Boernsman: I was pretty sure we agreed to invoke removeThing on reconfigure. If remove thing is not available for cleaning up, then I guess more plug-in are leaking currently.

homeconnect/integrationpluginhomeconnect.cpp
mzanetti: Hmm... I think this is still not ok...

So on a reconfigure the startPairing() is called again and in that case we already have a homeConnect object in the list which would be lost here, respectively in line 184 where this new one is inserted in the hash without cleaning up the old one... 

You do the check in setupThing() now and delete the existing one there...

So this seems like on a first-time setup, you create 2 of the homeConnect objects, one for pairing and one for operation. No leaks there, but also not needed to destroy the one from the pairing I guess...

On a reconfigure instead, it seems you're leaking the old one, and then additionally create 2 new ones (one for pairing, one for operation).

mzanetti (changes requested):

PR #277, opened 2020-06-15
Tado Plug-In: Fixed crash during thing pairing

Boernsman:

### DIFF
--- a/tado/integrationplugintado.cpp
+++ b/tado/integrationplugintado.cpp
@@ -61,7 +61,7 @@ void IntegrationPluginTado::startPairing(ThingPairingInfo *info)
 
 void IntegrationPluginTado::confirmPairing(ThingPairingInfo *info, const QString &username, const QString &password)
 {
-    qCDebug(dcTado()) << "Confirm pairing" << username << "password" << password << "Network manager available" << hardwareManager()->networkManager()->available();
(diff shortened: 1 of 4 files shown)


tado/integrationplugintado.cpp
```diff
@@ -56,13 +57,12 @@ void IntegrationPluginTado::startPairing(ThingPairingInfo *info)
             info->finish(Thing::ThingErrorNoError, QT_TR_NOOP("Please enter the login credentials for your Tado account."));
         }
     });
-    connect(reply, &QNetworkReply::finished, this, &QNetworkReply::deleteLater);
 }
 
 void IntegrationPluginTado::confirmPairing(ThingPairingInfo *info, const QString &username, const QString &password)
 {
+    qCDebug(dcTado()) << "Confirm pairing" << username << "password" << password << "Network manager available" << hardwareManager()->networkManager()->available();
```
mzanetti, line 64: Please don't print passwords in the logs... That makes it itchy to ask users for logs. What I normally do is this:

`password.replace(QRegex(".*","*"))`
to replace every character with a *, so one can see in the logs if a valid password has been set (i.e. just one * or none at all would hint to a bad login already) but not what the password itself is.

Boernsman: removed password from debug output

PR #313, opened 2020-09-04
New Plugin: Fastcom

mzanetti:


fastcom/integrationpluginfastcom.json
```diff
@@ -0,0 +1,83 @@
...
+                        {
+                            "id": "c29dcc25-40b1-4db8-a8ad-6552b2813ee8",
+                            "name": "testCompleted",
+                            "displayName": "Test run completed",
+                            "paramTypes": [
+                                {
+                                    "id": "3d101deb-a3b7-451c-ad07-02ff38afd617",
+                                    "name": "result",
+                                    "displayName": "Test result",
+                                    "type": "double",
```
Boernsman, line 37: A bit confusing without any unit. Mbps, MBs, kBs ..

Boernsman (changes requested):
Readme, meta.json, fastcom.png and translation files are missing

fastcom/integrationpluginfastcom.json
mzanetti: ah right... yeah... we're lacking a unit for that still...

mzanetti: added readme meta and icon

fastcom/integrationpluginfastcom.py
```diff
@@ -0,0 +1,62 @@
...
+    logger.log("thingRemoved:", thing.name)
+    timers[thing].cancel()
+    del timers[thing]
+
+
+def executeAction(info):
+    logger.log("executeAction")
+    if info.actionTypeId == speedtestRunTestActionTypeId:
+        info.finish(nymea.ThingErrorNoError)
+        runTest(info.thing, duration=info.paramValue(speedtestRunTestActionDurationParamTypeId))
```
t-mon, line 41: Maybe only run the test if currently no test is running, this could have an impact on both results. Using the state for that would be the easiest way. In case a test is already running, return no error I'd suggest.

t-mon (changes requested):

Boernsman (approved):

fastcom/integrationpluginfastcom.py
mzanetti: Fixed... I actually changed to return ThingErrorInUse and a error message that a test is already running which seems more correct IMO

mzanetti:
merged

PR #321, opened 2020-09-23
SolarLog: Fix units in json

mzanetti:

@boernsman, please check this

### DIFF
--- a/solarlog/integrationpluginsolarlog.json
+++ b/solarlog/integrationpluginsolarlog.json
@@ -48,7 +48,7 @@
                             "displayName": "Power AC",
                             "displayNameEvent": "Power AC changed",
                             "type": "double",
-                            "unit": "KiloWatt",
+                            "unit": "Watt",
                             "defaultValue": 0
                         },
                         {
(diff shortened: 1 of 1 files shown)


solarlog/integrationpluginsolarlog.json
```diff
@@ -48,7 +48,7 @@
                             "displayName": "Power AC",
                             "displayNameEvent": "Power AC changed",
                             "type": "double",
-                            "unit": "KiloWatt",
+                            "unit": "Watt",
```
Boernsman, line 51: Use proper unit ... in my point of view KiloWatt is the propper unit for this use cases, solarlog itself displays this values as well in KiloWatt. I guess this change was done because of a missmacht between unit (kW) and actual value (W)?

Boernsman (approved):
In my opinion kW is to be used.

solarlog/integrationpluginsolarlog.json
mzanetti: just noticed that the interface does not dictate a unit... That's actually the place where this is to be fixed.

Now for Watt vs KiloWatt... I don't even have a strong opinion on it. What is to be shown in the end to the user is really up to the client app... It should show Watts for values < 1000 and kW for larger values dynamically...

So in that sense what we should use in here is what is the most common across *all* smartmeters and probably it would make sense to keep consumer and producer the same in terms of units to not confuse the developers (plugin and client).

mzanetti: In here I changed it to Watts because that's what the code actually does...

PR #330, opened 2020-10-21
AQI: Added API key provider

Boernsman:

### DIFF
--- a/aqi/integrationpluginaqi.cpp
+++ b/aqi/integrationpluginaqi.cpp
@@ -38,50 +38,13 @@ IntegrationPluginAqi::IntegrationPluginAqi()
 
 }
 
-void IntegrationPluginAqi::startPairing(ThingPairingInfo *info)
-{
-    NetworkAccessManager *network = hardwareManager()->networkManager();
-    QNetworkReply *reply = network->get(QNetworkRequest(QUrl("https://api.waqi.info")));
-    connect(reply, &QNetworkReply::finished, this, [reply, info] {
-        reply->deleteLater();
-
-        if (reply->error() == QNetworkReply::NetworkError::HostNotFoundError) {
-            info->finish(Thing::ThingErrorHardwareNotAvailable, QT_TR_NOOP("Air quality index server is not reachable."));
-        } else {
-            info->finish(Thing::ThingErrorNoError, QT_TR_NOOP("Please enter your API token for Air Quality Index"));
-        }
-    });
-}
-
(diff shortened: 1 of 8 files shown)


aqi/integrationpluginaqi.cpp
```diff
@@ -30,63 +30,27 @@
...
-    connect(reply, &QNetworkReply::finished, this, [reply, info] {
-        reply->deleteLater();
-
-        if (reply->error() == QNetworkReply::NetworkError::HostNotFoundError) {
-            info->finish(Thing::ThingErrorHardwareNotAvailable, QT_TR_NOOP("Air quality index server is not reachable."));
-        } else {
-            info->finish(Thing::ThingErrorNoError, QT_TR_NOOP("Please enter your API token for Air Quality Index"));
+        if (paramTypeId == airQualityIndexPluginApiKeyParamTypeId && m_aqiConnection) {
+            if (!value.toString().isEmpty())
+                m_aqiConnection->setApiKey(value.toString());
```
mzanetti, line 43: uhmm... Wouldn't this crash if you set the API without actually having a AQI thing set up?

aqi/integrationpluginaqi.json
```diff
@@ -1,11 +1,21 @@
 {
     "name": "AirQualityIndex",
-    "displayName": "AirQualityIndex",
+    "displayName": "Air quality index",
     "id": "57d69b76-4d2d-41ec-bef6-949a79ffbe6b",
+    "paramTypes": [
+        {
+            "id": "b6861adb-7ed5-445f-b500-4df9eab866ef",
+            "name": "apiKey",
+            "displayName": "API key",
```
mzanetti, line 9: "Custom API key" would IMO be better to make it clearer that there is one that should work normally...

mzanetti (changes requested):
One issue spottet (I believe) and 2 tiny suggestions

aqi/integrationpluginaqi.cpp
Boernsman: No, since there is a check if m_aqiConnection is not a nullptr

mzanetti: Are you really sure about that?

How I read this code it seems to be a nullptr when neither setupThing() nor discoverThings() has been called, which might be the case on a fresh startup where no AQI things are configured.

mzanetti: stupid me... I missed the check for the valid aqiConnectiono in the if...

mzanetti (approved):
happroved

PR #344, opened 2020-11-04
Shelly: Add support for the Shelly 3EM

mzanetti:

### DIFF
--- a/shelly/integrationpluginshelly.json
+++ b/shelly/integrationpluginshelly.json
@@ -926,6 +926,290 @@
                     ]
                 },
                 {
+                    "id": "ba293550-d2af-4463-b973-e1812ab67b96",
+                    "name": "shellyEm3",
+                    "displayName": "Shelly 3EM",
+                    "createMethods": ["discovery"],
+                    "interfaces": [ "extendedsmartmeterconsumer", "extendedsmartmeterproducer", "wirelessconnectable", "update"],
+                    "paramTypes": [
+                        {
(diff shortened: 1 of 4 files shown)


shelly/integrationpluginshelly.json
```diff
@@ -924,6 +924,290 @@
...
+                            "readOnly": true
+                        },
+                        {
+                            "id": "21e11417-b862-44e6-828f-d65207328630",
+                            "name": "username",
+                            "displayName": "Username (optional)",
+                            "type": "QString"
+                        },
+                        {
+                            "id": "29aee8f6-ed40-41c5-83c9-32c8aa89c7ea",
```
Boernsman, line 948: Why not setup method userandpassword?

mzanetti: By default, Shelly devices don't use authentication. If a user/pass is set here, nymea will actually activate authentication on the device (so that from this point on, even the webinterface on the shelly will need this credentials). I suspect most users will not want to activate this. At this point our authentication mechanism can't deal with optional authentication.

We could think about supporting that in the future... For example that that startPairing checks if a device actually needs auth and allow the UI to skip the actual login if not needed. I think we needed that for another use case too. But it's not here yet.

Boernsman (approved):

shelly/integrationpluginshelly.json
mzanetti: Agreed to have the missing core feature soonish and then fix this properly

mzanetti:
merged

PR #355, opened 2020-11-29
Nuimo: Remove Battery Critical state

simonseres:
Removed the BatteryCriticalState since battery level is not conveyed right anyway.

### DIFF
--- a/senic/integrationpluginsenic.cpp
+++ b/senic/integrationpluginsenic.cpp
@@ -105,7 +105,6 @@ void IntegrationPluginSenic::setupThing(ThingSetupInfo *info)
     connect(nuimo, &Nuimo::rotationValueChanged, this, &IntegrationPluginSenic::onRotationValueChanged);
     connect(nuimo, &Nuimo::connectedChanged, this, &IntegrationPluginSenic::onConnectedChanged);
     connect(nuimo, &Nuimo::deviceInformationChanged, this, &IntegrationPluginSenic::onDeviceInformationChanged);
-    connect(nuimo, &Nuimo::batteryValueChanged, this, &IntegrationPluginSenic::onBatteryValueChanged);
 
     m_nuimos.insert(nuimo, thing);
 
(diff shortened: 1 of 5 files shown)


mzanetti:
just removing this state won't work as it's a mandatory state of the "battery" interface... Possible options are:
* Fix the state reporting (preferred option)
* Leave it there but never actually set it to true
* Completely remove the battery interface and all its related states.

Can you explain a bit more what's wrong with this? Is it triggering too early (i.e. does the nuimo still run for a long time even if critical is reported) or is it flickering on/off repeatedly?

simonseres:
> just removing this state won't work as it's a mandatory state of the "battery" interface... Possible options are:
got it
> 
> * Fix the state reporting (preferred option)
> * Leave it there but never actually set it to true
> * Completely remove the battery interface and all its related states.
thats the way to go
> 
> Can you explain a bit more what's wrong with this? Is it triggering too early (i.e. does the nuimo still run for a long time even if critical is reported) or is it flickering on/off repeatedly?

Battery is always shown at 0%. 
The battery status in the Nuimo plugin is overengineering anyways. Nuimo has this LED matrix that shows low battery. In practical life, there is no benefit of having this information in nymea.
And since Nuimo is not a popular device, roughly a handful of people use it with nymea. 
So it doesn't make sense spending more time on that issue (2 people spend some hours to make this work in the past).
I will remove the whole battery state then.

mzanetti:
I've read up on this and have understood the issue... This doesn't only affect the nuimo tho, it affects most bluetooth devices. Also this should be working fine on maveo boxes, just not on other setups... tbh I'd prefer fixing it also for our rpi images/other installations. Mainly because it does not only affect the nuimo...

For a test, could you please create this file on your machine:
`/etc/systemd/system/bluetooth.service.d/01-disable-unwanted-plugins.conf`
with the content
```
[Service]
ExecStart=
ExecStart=/usr/lib/bluetooth/bluetoothd --noplugin=battery,sap
```

and then reboot? It should make it work.

simonseres:
Created the file/content and rebooted.
Problem persists.
image

t-mon (changes requested):
Why do we want to remove the battery state? 
I understand that depending on the bluez version the battery has to be handled different, but is it not working any more?

mzanetti:
Closing this.

PR #407, opened 2021-03-24
New plugin: Neato botvac

mzanetti:

@loosrob tested it


neatobotvac/integrationpluginneatobotvac.py
```diff
@@ -0,0 +1,222 @@
...
+            logger.log("rbtBoundData Type: ", type(rbtBoundData), "rbtBoundData Contents: ", rbtBoundData)
+            thingDescriptor.params = [
+                nymea.Param(robotThingSerialParamTypeId, robot.serial),
+                nymea.Param(robotThingSecretParamTypeId, robot.secret),
+                nymea.Param(robotThingMapIdParamTypeId, mapId)
+            ]
+            thingDescriptors.append(thingDescriptor)
+
+        # And let nymea know about all the users robots
+        autoThingsAppeared(thingDescriptors)
```
loosrob, line 103: Does this have to be in the "for" loop to avoid calling it too much and creating the same robot multiple times?

mzanetti: No, that's fine outside... the parameter thingDescripts would be empty if no new robots are found. I think the issue is somewhere around line 72 where we check for a robot already existing... Perhaps some timing issue... Would be good to get the logs of such a run where it wrongly creates such an additional robot.

I also notice that ThingDescriptor is missing api still... It's not possible to add the parentId with the pyhton api. I'll have to add that.

loosrob: Attached some logs with following scenario's (after removing all Neato-things): add neato account & discovery of 1 robot, reboot & discovery of 2 identical robots. Does this help?

This one is strange:
 W | PythonIntegrations: No param for paramTypeId: QUuid("{def9a4bb-7a7e-4e3a-a63c-c55a105abb5e}")

This should be the robot serial, but all robots have (the same) serial number

[logs.docx](https://github.com/nymea/nymea-plugins/files/6265549/logs.docx)

mzanetti: Hmm, The PythonIntegrations logs aren't that interesting to debug this problem. Can you please enable logs for "Neato" and "ThingManager" and redo this?

loosrob: I tried 10 times or so, and I can't seem to reproduce this on nymead 0.27.0+202103192245~focal1. In any case, I attached the logs.

The logs in the previous post were from nymead 0.27.0+202103192245~buster+rpi1, where I was able to reproduce the error. The attached logs show that 2 robots were present when I started the tests, but after removing them, the plugin played nice and didn't add a 2nd robot anymore for about 10 restarts, but finally gave me 2 robots again but it gave an error and polling didn't start. After reboot both robots were still present, and polling worked.
[logs.docx](https://github.com/nymea/nymea-plugins/files/6270605/logs.docx)

PR #442, opened 2021-06-04
WakeOnLan: Add network device discovery 

t-mon:

### DIFF
--- a/wakeonlan/integrationpluginwakeonlan.cpp
+++ b/wakeonlan/integrationpluginwakeonlan.cpp
@@ -47,11 +47,11 @@ void IntegrationPluginWakeOnLan::discoverThings(ThingDiscoveryInfo *info)
 {
     if (!hardwareManager()->networkDeviceDiscovery()->available()) {
         qCWarning(dcWakeOnLan()) << "Failed to discover network devices. The network device discovery is not available.";
-        info->finish(Thing::ThingErrorHardwareNotAvailable, QT_TR_NOOP("The discovery is not available."));
+        info->finish(Thing::ThingErrorHardwareNotAvailable, QT_TR_NOOP("Unable to discovery devices in your network."));
         return;
     }
 
-    qCDebug(dcWakeOnLan()) << "Start discovering network devices...";
+    qCDebug(dcWakeOnLan()) << "Starting network discovery...";
     NetworkDeviceDiscoveryReply *discoveryReply = hardwareManager()->networkDeviceDiscovery()->discover();
     connect(discoveryReply, &NetworkDeviceDiscoveryReply::finished, this, [=](){
         ThingDescriptors descriptors;
--- a/wakeonlan/integrationpluginwakeonlan.cpp
+++ b/wakeonlan/integrationpluginwakeonlan.cpp
@@ -69,7 +69,7 @@ void IntegrationPluginWakeOnLan::discoverThings(ThingDiscoveryInfo *info)
             if (networkDevice.hostName().isEmpty()) {
                 title = networkDevice.address().toString();
             } else {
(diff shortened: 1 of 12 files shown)


wakeonlan/integrationpluginwakeonlan.cpp
```diff
@@ -41,19 +43,64 @@ IntegrationPluginWakeOnLan::IntegrationPluginWakeOnLan()
 {
 }
 
+void IntegrationPluginWakeOnLan::discoverThings(ThingDiscoveryInfo *info)
+{
+    if (!hardwareManager()->networkDeviceDiscovery()->available()) {
+        qCWarning(dcWakeOnLan()) << "Failed to discover network devices. The network device discovery is not available.";
+        info->finish(Thing::ThingErrorHardwareNotAvailable, QT_TR_NOOP("The discovery is not available."));
```
mzanetti, line 50: This isn't a very user friendly error message... IMO something like "Unable to discovery devices in your network." would be better for an end user. This one is fine for a warning message in the logs.

wakeonlan/integrationpluginwakeonlan.cpp
```diff
@@ -41,19 +43,64 @@ IntegrationPluginWakeOnLan::IntegrationPluginWakeOnLan()
...
+        return;
+    }
+
+    qCDebug(dcWakeOnLan()) << "Start discovering network devices...";
```
mzanetti, line 54: "Starting network discovery..."

wakeonlan/integrationpluginwakeonlan.cpp
```diff
@@ -41,19 +43,64 @@ IntegrationPluginWakeOnLan::IntegrationPluginWakeOnLan()
...
+    NetworkDeviceDiscoveryReply *discoveryReply = hardwareManager()->networkDeviceDiscovery()->discover();
+    connect(discoveryReply, &NetworkDeviceDiscoveryReply::finished, this, [=](){
+        ThingDescriptors descriptors;
+        qCDebug(dcWakeOnLan()) << "Discovery finished. Found" << discoveryReply->networkDeviceInfos().count() << "devices";
+        foreach (const NetworkDeviceInfo &networkDeviceInfo, discoveryReply->networkDeviceInfos()) {
+            // We need the mac address...
+            if (networkDeviceInfo.macAddress().isEmpty())
+                continue;
+
+            // Filter out already added network devices, rediscovery is in this case no option
```
mzanetti, line 64: Hmm... Why preventing the rediscovery? I agree it doesn't make sense, but the API allows to do it so IMO it should return something sensible.

mzanetti: Would be enough to change the code like this I guess:

```
Thing *existingThing = myThings().findByParam(Param(wolThingMacParamTypeId, networkDeviceInfo.macAddress());
if (existingThing) {
    descriptor.setThingId(existingThing->id());
}
```

mzanetti (changes requested):
Minor nitpicks. Mostly fine.

wakeonlan/integrationpluginwakeonlan.cpp
t-mon: The idea was: if you reconfigure this thing, you probably wanna change the mac address, if not, add a new one, yes, but a rediscovery and showing only the one you already added makes no sense, that's why I didn't wanted to set the existing id...

mzanetti: I see it mostly like a "Hmm... something didn't seem to work, lemme try reconfiguring it". And then the user goes to reconfigure and reconfigure doesn't work either...

PR #500, opened 2021-10-30
Garadget: Initial version Garage Door Opener

hi-flyer:
Tested on Ubuntu 20.04 server with Garadget v1.24 Device and DECKO Garage Door Opener

Updated README

Updated the translations file

I do not know the correct answer to this question but it is a new plugin. 
I guess it should be nymea-plugins-all


hi-flyer:
Have updated system based on guidance given me. Will create a new pull-request based on the update

PR #502, opened 2021-10-31
New plugin: Garadget

hi-flyer:

- [x] Did you test the changes on hardware, if not (e.g. absence of required hardware), please mention a person to confirm it has been tested. Yes

- [x] Did you update the plugin's README.md accordingly? Yes and modified as suggested

- [x] Did you update translations (`cd builddir && make lupdate`)? yes for garadget only this time!

- [x] If you added a new plugin, should it be added to nymea-plugins-all or nymea-plugins-maker? I do not know the correct answer for this one.
- [x] git reset suggested did not remove all unwanted commits. I deleted the branch and recreated it adding all updates. So this should be the modified version for the plugin.

### DIFF
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -213,4 +213,10 @@ void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByt
         thing->setStateValue(garadgetMttStateTypeId,jo.value(QString("mtt")).toInt()/1000);
         qCDebug(dcGaradget) << "System Configuration" << "srt =" << thing->stateValue(garadgetSrtStateTypeId).toInt() << "rlt =" << thing->stateValue(garadgetRltStateTypeId).toInt()<< "mtt =" << thing->stateValue(garadgetMttStateTypeId).toInt() * 1000;
     }
+    if (topic.endsWith("/set-config")){
+        if ( (payload.contains("mqip"))  or (payload.contains("mqpt")) ) {
+            thing->setStateValue(garadgetConnectedStateTypeId, false);
+            qCDebug(dcGaradget) << "set connected to false";
+        }
+    }
 }
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -83,7 +83,7 @@ void IntegrationPluginGaradget::postSetupThing(Thing *thing)
         }
         name = name + "/command";
         qCDebug(dcGaradget) << "inside m_pluginTimer with" << name ;
-        uint updatetime = 10;
+        uint updatetime = 30;
         m_pluginTimer = hardwareManager()->pluginTimerManager()->registerTimer(updatetime);
         connect(m_pluginTimer, &PluginTimer::timeout, this, [=](){
             if (m_garadgetconnect == 1) {
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -106,7 +106,7 @@ void IntegrationPluginGaradget::thingRemoved(Thing *thing)
 {
     qCDebug(dcGaradget) << thing << "Removed";
     m_mqttClients.take(thing)->deleteLater();
-    if (m_pluginTimer && !myThings().isEmpty()) {
+    if (m_pluginTimer && myThings().isEmpty()) {
         hardwareManager()->pluginTimerManager()->unregisterTimer(m_pluginTimer);
         m_pluginTimer = nullptr;
     }
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -69,33 +69,30 @@ void IntegrationPluginGaradget::setupThing(ThingSetupInfo *info)
         qCDebug(dcGaradget) << "entered is Connected" << client;
         subscribe(thing);
     }
-    m_lastActivityTimeStamps[thing] = QDateTime::currentDateTime();
 }
 
 
 void IntegrationPluginGaradget::postSetupThing(Thing *thing)
 {
 
-    if (!m_pluginTimer) {
+    if (!m_pluginTimer && !myThings().isEmpty()) {
         QString name = thing->paramValue(garadgetThingDeviceNameParamTypeId).toString();
         if (name.endsWith("/#")) {
             name.chop(2);
         }
         name = name + "/command";
         qCDebug(dcGaradget) << "inside m_pluginTimer with" << name ;
-        uint updatetime = 30;
+        uint updatetime = 10;
         m_pluginTimer = hardwareManager()->pluginTimerManager()->registerTimer(updatetime);
         connect(m_pluginTimer, &PluginTimer::timeout, this, [=](){
-            if (m_garadgetconnect == 1) {
-                foreach (Thing *thing, myThings()) {
+            foreach (Thing *thing, myThings()) {
+                if ((m_lastActivityTimeStamps[thing].msecsTo(QDateTime::currentDateTime()) > 1000 * updatetime) && (thing->stateValue(garadgetConnectedStateTypeId).toBool() == true)) {
+                    qCDebug(dcGaradget) << "disconnect garadget";
+                    thing->setStateValue(garadgetConnectedStateTypeId, false);
+                }
+                if (thing->stateValue(garadgetConnectedStateTypeId).toBool() == true) {
                     m_mqttClients.value(thing)->publish(name, "get-status");
                 }
-            }
-            uint timesinceupdate = QDateTime::currentDateTime().toTime_t() - m_lastActivityTimeStamps[thing].toTime_t();
-            if ((timesinceupdate > updatetime) && (m_garadgetconnect == 1)) {
-                qCDebug(dcGaradget) << "disconnect garadget" << m_lastActivityTimeStamps[thing] << timesinceupdate ;
-                thing->setStateValue(garadgetConnectedStateTypeId, false);
-                m_garadgetconnect = 0;
             }
         });
     }
@@ -207,20 +204,19 @@ void IntegrationPluginGaradget::subscribe(Thing *thing)
 
 void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByteArray &payload, bool retained)
 {
-    qCDebug(dcGaradget) << "Received message from topic" << topic << "with msg" << payload << "retain flag" << retained;
+//    qCDebug(dcGaradget) << "Received message from topic" << topic << "with msg" << payload << "retain flag" << retained;
 
 
     MqttClient* client = static_cast<MqttClient*>(sender());
     Thing *thing = m_mqttClients.key(client);
     if (!thing) {
-        qCWarning(dcGaradget) << "Received a publish message from a client where de don't have a matching thing";
+        qCWarning(dcGaradget) << "Received a publish message from a client where de don't have a matching thing" << retained;
         return;
     }
     if (topic.endsWith("/status")) {
         if (thing->stateValue(garadgetConnectedStateTypeId) == false) {
             qCDebug(dcGaradget) << "Setting Garadget to connected" ;
             thing->setStateValue(garadgetConnectedStateTypeId, true);
-            m_garadgetconnect = 1;
         }
         m_lastActivityTimeStamps[thing] = QDateTime::currentDateTime();
         QJsonParseError error;
@@ -230,7 +226,7 @@ void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByt
             return;
         }
         QJsonObject jo = jsonDoc.object();
-        thing->setStateValue(garadgetSignalStrengthStateTypeId, (100 + jo.value(QString("signal")).toInt()) * 2 );
+        thing->setStateValue(garadgetSignalStrengthStateTypeId, (99 + jo.value(QString("signal")).toInt()) * 2 );
         thing->setStateValue(garadgetSensorlevelStateTypeId, jo.value(QString("sensor")).toInt());
         thing->setStateValue(garadgetBrightlevelStateTypeId, jo.value(QString("bright")).toInt());
         if (jo.value(QString("status")).toString().contains(QString("stopped"))) {
@@ -256,7 +252,6 @@ void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByt
     if (topic.endsWith("/set-config")){
         if ( (payload.contains("mqip"))  or (payload.contains("mqpt")) ) {
             thing->setStateValue(garadgetConnectedStateTypeId, false);
-            m_garadgetconnect = 0;
             qCDebug(dcGaradget) << "Detected change of Broker msg - set connected to false";
         }
     }
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -141,7 +141,7 @@ void IntegrationPluginGaradget::executeAction(ThingActionInfo *info)
     QString conftype = "";
     int actint = 0;
     if (action.actionTypeId() == garadgetSrtActionTypeId) {
-        if ( (action.paramValue(garadgetSrtActionSrtParamTypeId).toInt() > -1) and (action.paramValue(garadgetSrtActionSrtParamTypeId).toInt() < 81)) {
+        if (action.paramValue(garadgetSrtActionSrtParamTypeId).toInt() > -1) {
             actint = action.paramValue( garadgetSrtActionSrtParamTypeId).toInt();
             conftype = "srt";
         } else {
@@ -150,8 +150,8 @@ void IntegrationPluginGaradget::executeAction(ThingActionInfo *info)
         }
     }
     if (action.actionTypeId() == garadgetMttActionTypeId) {
-        if ( (action.paramValue( garadgetMttActionMttParamTypeId).toInt() > 4) and (action.paramValue( garadgetMttActionMttParamTypeId).toInt() < 61) ){
-            actint = action.paramValue( garadgetMttActionMttParamTypeId).toInt() * 1000;
+        if (action.paramValue( garadgetMttActionMttParamTypeId).toInt() > 0) {
+            actint = action.paramValue( garadgetMttActionMttParamTypeId).toInt();
             conftype = "mtt";
         } else {
             name = name + "/command";
@@ -159,7 +159,7 @@ void IntegrationPluginGaradget::executeAction(ThingActionInfo *info)
         }
     }
     if (action.actionTypeId() == garadgetRltActionTypeId) {
-        if ( (action.paramValue( garadgetRltActionRltParamTypeId).toInt() > 9) and (action.paramValue( garadgetRltActionRltParamTypeId).toInt() < 2001) ){
+        if (action.paramValue( garadgetRltActionRltParamTypeId).toInt() > 0) {
             actint = action.paramValue(garadgetRltActionRltParamTypeId).toInt();
             conftype = "rlt";
         } else {
@@ -246,7 +246,7 @@ void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByt
         QJsonObject jo = jsonDoc.object();
         thing->setStateValue(garadgetSrtStateTypeId,jo.value(QString("srt")).toInt());
         thing->setStateValue(garadgetRltStateTypeId,jo.value(QString("rlt")).toInt());
-        thing->setStateValue(garadgetMttStateTypeId,jo.value(QString("mtt")).toInt()/1000);
+        thing->setStateValue(garadgetMttStateTypeId,jo.value(QString("mtt")).toInt());
         qCDebug(dcGaradget) << "System Configuration" << "srt =" << thing->stateValue(garadgetSrtStateTypeId).toInt() << "rlt =" << thing->stateValue(garadgetRltStateTypeId).toInt()<< "mtt =" << thing->stateValue(garadgetMttStateTypeId).toInt() * 1000;
     }
     if (topic.endsWith("/set-config")){
--- a/garadget/integrationplugingaradget.cpp
+++ b/garadget/integrationplugingaradget.cpp
@@ -80,7 +80,7 @@ void IntegrationPluginGaradget::postSetupThing(Thing *thing)
         if (name.endsWith("/#")) {
             name.chop(2);
         }
-        name = name + "/command";
+
         qCDebug(dcGaradget) << "inside m_pluginTimer with" << name ;
         uint updatetime = 30;
         m_pluginTimer = hardwareManager()->pluginTimerManager()->registerTimer(updatetime);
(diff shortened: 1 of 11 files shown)


garadget/integrationplugingaradget.json
```diff
@@ -0,0 +1,184 @@
...
+                            "id": "174509b5-fc3e-4f03-8c93-8e044f0a2ce0" ,
+                            "name": "state",
+                            "displayName": "State",
+                            "displayNameEvent": "State changed",
+                            "type": "QString",
+                            "possibleValues": ["open", "closed", "opening", "closing", "intermediate"],
+                            "defaultValue": "closed"
+                        },
+                        {
+                            "id": "024f178c-a920-42d4-887f-1c90a96d5eb3",
```
mzanetti, line 48: If you add the "wirelessconnectable" interface and rename this state to "signalStrength", nymea:app will show the signal strength icon in the UI.

garadget/integrationplugingaradget.json
```diff
@@ -0,0 +1,184 @@
...
+                            "id": "4aaeefbd-46d9-4111-b262-a001a47ecb22",
+                            "name": "sensorlevel",
+                            "displayName": "Sensor Level",
+                            "displayNameEvent": "Sensor Level changed",
+                            "type": "int",
+                            "defaultValue": 0
+                        },
+                        {
+                            "id": "bd7fece8-3ff6-421a-bb5e-80323beb8d8d",
+                            "name": "brightlevel",
```
mzanetti, line 65: Also here you can add the "lightsensor" interface and rename the state to "lightIntensity" and nymea:app will show it in the sensors section for a light sensor too.

garadget/integrationplugingaradget.json
```diff
@@ -0,0 +1,184 @@
...
+                            "id": "024f178c-a920-42d4-887f-1c90a96d5eb3",
+                            "name": "signallevel",
+                            "displayName": "Signal Level",
+                            "displayNameEvent": "Signal Level changed",
+                            "type": "int",
+                            "defaultValue": 0
+                        },
+                        {
+                            "id": "4aaeefbd-46d9-4111-b262-a001a47ecb22",
+                            "name": "sensorlevel",
```
mzanetti, line 57: What sort of sensor is this?

garadget/integrationplugingaradget.json
```diff
@@ -0,0 +1,184 @@
...
+                            "displayNameEvent": "Reflection Threshold changed",
+                            "displayNameAction": "Set Reflection Threshold value",
+                            "type": "int",
+                            "defaultValue": 25,
+                            "writable": true
+                        },
+                        {
+                            "id": "c38e89db-b259-4fa0-9f20-03fe3b8a7514" ,
+                            "name": "rlt",
+                            "displayName": "Button Press Time in 10-2000 ms",
```
mzanetti, line 85: Wouldn't this make more sense as a settingTypes instead of a stateTypes?

garadget/integrationplugingaradget.json
```diff
@@ -0,0 +1,184 @@
...
+                            "displayNameEvent": "Button Press Time changed",
+                            "displayNameAction": "Set Button Press Time value",
+                            "type": "int",
+                            "defaultValue": 300,
+                            "writable": true
+                        },
+                        {
+                            "id": "acda9268-4663-46d1-a409-231d648e6fc8" ,
+                            "name": "mtt",
```
mzanetti, line 94: Same here, seems to be more of a setting instead of a state, no?

mzanetti:
@nymea-jenkins test this please

hi-flyer:
so what was the correct answer to this one? should it be added to nymea-plugins-all or nymea-plugins-maker?

mzanetti:
> so what was the correct answer to this one? should it be added to nymea-plugins-all or nymea-plugins-maker?

yes, to nymea-plugins-all

garadget/integrationplugingaradget.json
hi-flyer: I did implement this but am rejecting as I do not see a good reason to implement. I did not see the icon you talk about and further, this assumes a percentage level. while I can trick it to a number between 0 and 100, it is really an rssi value if I understand it correctly. It is there, in my opinion to help the user know that the device has a good connection to the wifi access point. Also I did not understand how I could set this value when received from garadget.

garadget/integrationplugingaradget.json
hi-flyer: The light sensor mentioned above is used to detect the lazed beam reflected off of the garage door or wall depending on where you put the reflector. Modification of the value associate with this level helps eliminate false detected state changes. the value of this does change depending on the value of the brightness level. so to summarize it is a calculated value of the reflected lazer signal level. I did not read this anywhere but that is what it is.

garadget/integrationplugingaradget.json
hi-flyer: tried to modify this but did not see a way to set it based on the returned values from the garadget device. Ok, I struggle with QT and did not find anything on how to set the value if changed to settingsType.

garadget/integrationplugingaradget.json
hi-flyer: same as above, tried to modify this but did not see a way to set it based on the returned values from the garadget device. Ok, I struggle with QT and did not find anything on how to set the value if changed to settingsType.

hi-flyer:
responded to mzanetti suggestions above.

garadget/integrationplugingaradget.json
hi-flyer: I have made the change in my test unit but do not see any change in the UI.

garadget/integrationplugingaradget.json
hi-flyer: I am change the display name to Laser Sensor Level to help user understand.

mzanetti:
Fails the build. `garadget` is missing in nymea-plugins.pro so it won't be built and the packaging complains about it being missing,

garadget/integrationplugingaradget.json
mzanetti: rssi to percentage:
percentage = (rssi + 50) / 100

mzanetti: About the icon you're not seeing: That really must be working, it's being used by tons of other plugins. Can you show the changes you made? there must be a little mistake somewhere.

garadget/integrationplugingaradget.json
mzanetti: oh, I see. Yes, in this case it's not an acutal lightSensor. Fine to not use that interface for this then.

garadget/integrationplugingaradget.json
mzanetti: You can call `thing->setSettingsValue(<typeId>, <value>)` to update the settings value programmatically.

garadget/integrationplugingaradget.json
mzanetti: thing->setSettingsType()

hi-flyer:
Have installed experimental on my Ubuntu 20.04 version. When I try to activate it, it shows uh oh, something when wrong. I compilied my latest version, that I uploaded this morning on Ubuntu 20.04 and placed libnymea_integrationplugingaradget.so on the operational system and it works as designed. not sure what the problem is with the experimental

mzanetti:
> not sure what the problem is with the experimental

The package in the experimental has a build timestamp in the version. It will only include your changes that have been pushed before that time (It's in UTC). So depending on when you pushed your changes and when we've triggered a build for the repository, it may or may not contain your latest changes.

That said, checking the logs during the setup should also help with finding out what's going on.

hi-flyer:
I have removed garadget and reinstalled but now the garadget does not show up on the config menu. tried multiple times. The Nov 2 version (according to the date on the file) is what it installs. nothing showing in the nymea logs from the main log viewer.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+        QString name = thing->paramValue(garadgetThingDeviceNameParamTypeId).toString();
+        if (name.endsWith("/#")) {
+            name.chop(2);
+        }
+        name = name + "/command";
+        qCDebug(dcGaradget) << "inside m_pluginTimer with" << name ;
+        uint updatetime = 30;
+        m_pluginTimer = hardwareManager()->pluginTimerManager()->registerTimer(updatetime);
+        connect(m_pluginTimer, &PluginTimer::timeout, this, [=](){
+            if (m_garadgetconnect == 1) {
```
mzanetti, line 89: I don't understand why this if(m_garadgetconnect == 1). Seems like the timer should publish the get-status unconditionally to every configured garadget system, no?

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+                foreach (Thing *thing, myThings()) {
+                    m_mqttClients.value(thing)->publish(name, "get-status");
+                }
+            }
+            uint timesinceupdate = QDateTime::currentDateTime().toTime_t() - m_lastActivityTimeStamps[thing].toTime_t();
+            if ((timesinceupdate > updatetime) && (m_garadgetconnect == 1)) {
```
mzanetti, line 95: A leaner way of checking how much time passed would be:
```
if (m_lastiActivityTimeStamps[thing].msecsTo(QDateTime::currentDateTime()) > updateTime) { ... }
```

Again, not sure why the m_garadgetconnect == 1

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+                qCDebug(dcGaradget) << "disconnect garadget" << m_lastActivityTimeStamps[thing] << timesinceupdate ;
+                thing->setStateValue(garadgetConnectedStateTypeId, false);
+                m_garadgetconnect = 0;
```
mzanetti, line 98: The biggest issue with the m_garadgetconnect is that you only store one. If a user has multiple garadgets connected, this will break because the two (or more) will overwrite each other's m_garadgetconnect.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+    Thing *thing = m_mqttClients.key(client);
+    if (!thing) {
+        qCWarning(dcGaradget) << "Received a publish message from a client where de don't have a matching thing";
+        return;
+    }
+    if (topic.endsWith("/status")) {
+        if (thing->stateValue(garadgetConnectedStateTypeId) == false) {
+            qCDebug(dcGaradget) << "Setting Garadget to connected" ;
+            thing->setStateValue(garadgetConnectedStateTypeId, true);
+            m_garadgetconnect = 1;
```
mzanetti, line 223: if this is just about memorizing the connected status, you might as well read the things statevalue with

```
thing->stateValue(garadgetConnectedStateTypeId).toBool()
```
This will return what you've last set to it. I.e. it will be in sync with your m_garadgetconnect variable and has the additional benefit that it will also work with multiple.

Anyhow, still not sure why it's needed at all.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+    connect(client, &MqttClient::subscribeResult, info, [info](quint16 /*packetId*/, const Mqtt::SubscribeReturnCodes returnCodes){
+        info->finish(returnCodes.first() == Mqtt::SubscribeReturnCodeFailure ? Thing::ThingErrorHardwareFailure : Thing::ThingErrorNoError);
+    });
+    connect(client, &MqttClient::publishReceived, this, &IntegrationPluginGaradget::publishReceived);
+    // In case we're already connected, manually call subscribe now
+    if (client->isConnected()) {
+        qCDebug(dcGaradget) << "entered is Connected" << client;
+        subscribe(thing);
+    }
+    m_lastActivityTimeStamps[thing] = QDateTime::currentDateTime();
```
mzanetti, line 72: I wouldn't even initialize it here. Because when the system starts up, you haven't ever heard from the device, you don't know if it's connected so it's wise to assume it is not connected at this point.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+
+
+void IntegrationPluginGaradget::thingRemoved(Thing *thing)
+{
+    qCDebug(dcGaradget) << thing << "Removed";
+    m_mqttClients.take(thing)->deleteLater();
+    if (m_pluginTimer) {
+        hardwareManager()->pluginTimerManager()->unregisterTimer(m_pluginTimer);
+        m_pluginTimer = nullptr;
```
mzanetti, line 112: Generally, good that you've thought of cleaning up. However, current code suffers from the issue that if you add 2 garadget devices, and then remove one of them, the timer will be removed even though it would still be needed.

Easy fix: just add this to the if statement:
```
if (m_pluginTimer && myThings().isEmpty())
```

garadget/integrationplugingaradget.cpp
hi-flyer: if I get a disconnect, it continues to publish a get-status which It should not do if garadget is not on line. Garadget publishes it status when it comes online automatically.

hi-flyer: this gets into your later comment on m_garadgetconnect. It needs to check the status of each garadget. will work on.

hi-flyer: have removed m_garadgetconnect an use the state now as suggested below. Still believe that it should not request status if it is not online. so moved if inside the foreach.

garadget/integrationplugingaradget.cpp
hi-flyer: m_garadgetconnect is removed

garadget/integrationplugingaradget.cpp
hi-flyer: m_garadgetconnect is now gone. I assume this solves your comment.

garadget/integrationplugingaradget.cpp
hi-flyer: accepted. it is now gone

garadget/integrationplugingaradget.cpp
hi-flyer: needed !myThing().isEmpty()
done

garadget/integrationplugingaradget.json
hi-flyer: reflection threshold and the sensor level are related. Want to see them on the same page from a user prospective. so not making this suggested change.

garadget/integrationplugingaradget.json
hi-flyer: note that reflection threshold needs to be on the same page as the Lazer sensor level. moving this to the settings page does not make sense in this case to me.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+#include "network/mqtt/mqttprovider.h"
+#include <mqttclient.h>
+
+
+void IntegrationPluginGaradget::setupThing(ThingSetupInfo *info)
+{
+    Thing *thing = info->thing();
+
+    QString device = thing->paramValue(garadgetThingDeviceNameParamTypeId).toString();
+    if (!(device.startsWith( "garadget/"))) {
```
mzanetti, line 46: To be hones I would just assume the user does *not* add "garadget/" and automatically do that in here... AFAIK it's also just "deviceName" in the garadget's settings, so if nymea asks for the device name, why would someone add "garadget/" to it?

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+        if (device.contains("/")) {
+            return info->finish(Thing::ThingErrorInvalidParameter, QT_TR_NOOP(QString ("The given deviceName is not valid - no /s allowed")));
+        }
+        thing->setParamValue(garadgetThingDeviceNameParamTypeId,"garadget/" + device + "/#" );
```
mzanetti, line 50: It's kinda bad practice to overwrite the parameters the user has set... Yes, it's possible, but if not absolutely required, we try to avoid that as it is confusing for the user.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    qCDebug(dcGaradget) << "entered setupThing" << thing->paramValue(garadgetThingDeviceNameParamTypeId) ;
+    MqttClient *client = nullptr;
+    if (thing->thingClassId() == garadgetThingClassId) {
```
mzanetti, line 55: why this if? Seems it's always true. Could just go aways IMO

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+}
+
+void IntegrationPluginGaradget::subscribe(Thing *thing)
+{
+    MqttClient *client = m_mqttClients.value(thing);
+    if (!client) {
+        // Device might have been removed
+        return;
+    }
+    if (thing->thingClassId() == garadgetThingClassId) {
```
mzanetti, line 216: again, the if here doesn't seem to be necessary. Given there is only one ThingClass in this plugin, this if will always be true. Can just go away

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+        client->subscribe(thing->paramValue(garadgetThingDeviceNameParamTypeId).toString());
```
mzanetti, line 217: I'd personally do:
```
client->subscribe("garadget/" + thing->paramValue(garadgetThingDeviceNameParamTypeId).toString());
```
As I said before, I'd assume the "garadget/" is not given by the user and would just always prepend it here and be done with it.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByteArray &payload, bool retained)
+{
+//    qCDebug(dcGaradget) << "Received message from topic" << topic << "with msg" << payload << "retain flag" << retained;
+
+
+    MqttClient* client = static_cast<MqttClient*>(sender());
+    Thing *thing = m_mqttClients.key(client);
+    if (!thing) {
+        qCWarning(dcGaradget) << "Received a publish message from a client where de don't have a matching thing" << retained;
```
mzanetti, line 229: typo :) "de" -> "we"

garadget/integrationplugingaradget.cpp
mzanetti: This still seems to be a problem. Creating 2 garadget things, and then removing one of them, will break the other

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+        qCDebug(dcGaradget) << "entered is Connected" << client;
+        subscribe(thing);
+    }
+}
+
+
+void IntegrationPluginGaradget::postSetupThing(Thing *thing)
+{
+
+    if (!m_pluginTimer && !myThings().isEmpty()) {
```
mzanetti, line 78: myThings() will never be empty in a postSetupThing() call. It will always contain at least the one thing that just has been set up.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+        QString name = thing->paramValue(garadgetThingDeviceNameParamTypeId).toString();
+        if (name.endsWith("/#")) {
+            name.chop(2);
+        }
+
+        qCDebug(dcGaradget) << "inside m_pluginTimer with" << name ;
```
mzanetti, line 84: Perhaps clean the debug messages a bit up. It's fine to have debug messages, but please make them nice texts and contain informative stuff.

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+                QJsonObject garadgetobj;
+
+                if ((thing->stateValue(garadgetConnectedStateTypeId).toBool() == true) && (settingTypeId == garadgetSettingsRdtParamTypeId)){
+                    garadgetobj.insert("rdt", value.toInt());
+                    garadgetobj.insert("rlp", thing->setting(garadgetSettingsRlpParamTypeId).toInt());
+                    QJsonDocument garadgetdoc(garadgetobj);
+                    QByteArray garadgetdata = garadgetdoc.toJson(QJsonDocument::Compact);
+                    QString jsonDoc(garadgetdata);
+                    qCDebug(dcGaradget()) << "Changing Configuration" << garadgetdata;
+                    m_mqttClients.value(thing)->publish(name + "/set-config", garadgetdata);
```
mzanetti, line 109: oh, obviously, if you follow my other suggestion about not assuiming the the "garadget/" in the name, then the publish calls would also need to have `"garadget/" + ` prepended. I suppose that's why you added it to the param in the first place, so you don't have to type it every time. Well, nymea philosophy is: The developers goes the extra mile if it's for a better user experience :)

garadget/integrationplugingaradget.cpp
```diff
@@ -0,0 +1,276 @@
...
+        actarray = act.toUtf8();
+        qCDebug(dcGaradget) << "Publishing:" << name << act;
+        quint16 packetId = client->publish(name, actarray, Mqtt::QoS1, false);
+        connect(client, &MqttClient::published, info, [info, packetId](quint16 packetIdResult){
+            if (packetId == packetIdResult) {
+                info->finish(Thing::ThingErrorNoError);
+            }
+        });
+    }
+    return;
```
mzanetti, line 206: Hmm... I can see paths where you end up in this return without having called info->finish(), no? What if there's a srcAction but the param is <= -1 ? "act" wil be empty and info->finish will never be called, no?
This will cause the ui in the app to spin for 30 seconds, waiting on the action to finish and then causing a timeout error.

mzanetti: Also, could it perhaps make sense to define "minValue" and "maxValue" for those srt states in the json file?

mzanetti (changes requested):

garadget/integrationplugingaradget.cpp
hi-flyer: Ok.

garadget/integrationplugingaradget.cpp
hi-flyer: I am aware of the 30 second timeout. was not sure it was appropriate to put info->finish(Thing::ThingErrorNoError); at the end but it is now there. I also point out that if -1 is entered the software will instead send a request for status will reset the user input to what is in the garadget. So I accepted you suggestion.

as far as minValue and maxValue, I did have them in there but that puts a slider up for which it was extremely hard to select the value I was want to enter. therefore I eliminated them to achieve the ability to enter the number I really wanted.

garadget/integrationplugingaradget.cpp
hi-flyer: also requires fix on other places where I took advantage of rename of nameDevice. am changing all places.

garadget/integrationplugingaradget.cpp
hi-flyer: I meant to remove. It was for me and forgot to remove. gone now.

garadget/integrationplugingaradget.cpp
hi-flyer: just part of not changing deviceName. accepted but you missed adding "/#" at the end.

garadget/integrationplugingaradget.cpp
hi-flyer: accepted

garadget/integrationplugingaradget.cpp
hi-flyer: accepted

garadget/integrationplugingaradget.cpp
hi-flyer: ok, am fixing all places where this affects

garadget/integrationplugingaradget.cpp
hi-flyer: sorry, but I accommodated one of your previous comments on this to do this but in the wrong spot. now put in the right spot. so comment accepted.

garadget/integrationplugingaradget.cpp
hi-flyer: this is where I placed  && !myThings().isEmpty() in the wrong spot. I should have put in the thingRemoved section.

accepted

mzanetti (approved):
LGTM :+1: 

Thanks for the contribution. Will be included in nymea 0.30.

hi-flyer:
Longer term testing shows false detection of unexpected disconnects. Testing a solution to resolve this issue.

garadget/integrationplugingaradget.cpp
```diff
@@ -255,4 +253,22 @@ void IntegrationPluginGaradget::publishReceived(const QString &topic, const QByt
             qCDebug(dcGaradget) << thing->paramValue(garadgetThingDeviceNameParamTypeId).toString() << "Detected change of Broker msg - set connected to false";
         }
     }
+    if (topic.endsWith("/LWT")){
+        if (payload.contains("Online") or payload.contains("Offline")) {
+            thing->setStateValue(garadgetLWTStateTypeId,true);
```
mzanetti, line 258: I don't think this should be a thing state... it's not of any interest to the user but only an implementation detail. You can store that in the pluginStorage instead:

```
pluginStorage()->beginGroup(thing->id());
pluginStorage()->setValue("lastWillAvailable", true);
pluginStorage()->endGroup();
```

and to read:
```
pluginStorage()->beginGroup(thing->id());
bool lastWillAvailable = pluginStorage()->setValue("lastWillAvailable").toBool();
pluginStorage()->endGroup();
```

mzanetti (changes requested):

garadget/integrationplugingaradget.cpp
hi-flyer: Agree and have made changes

mzanetti (approved):

PR #506, opened 2021-11-08
New Plugin: Zigbee remotes

mzanetti:

### DIFF
--- a/zigbeeremotes/integrationpluginzigbeeremotes.cpp
+++ b/zigbeeremotes/integrationpluginzigbeeremotes.cpp
@@ -1,0 +1,291 @@
+﻿/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+*
+*
+* Copyright 2013 - 2020, nymea GmbH
+* Contact: contact@nymea.io
+
+* This file is part of nymea.
+* This project including source code and documentation is protected by
+* copyright law, and remains the property of nymea GmbH. All rights, including
+* reproduction, publication, editing and translation, are reserved. The use of
+* this project is subject to the terms of a license agreement to be concluded
+* with nymea GmbH in accordance with the terms of use of nymea GmbH, available
+* under https://nymea.io/license
+*
+* GNU Lesser General Public License Usage
+* Alternatively, this project may be redistributed and/or modified under the
+* terms of the GNU Lesser General Public License as published by the Free
+* Software Foundation; version 3. This project is distributed in the hope that
+* it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+* warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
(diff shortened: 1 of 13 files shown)


zigbeeremotes/integrationpluginzigbeeremotes.cpp
```diff
@@ -0,0 +1,288 @@
...
+        ZigbeeNodeEndpoint *endpoint = node->getEndpoint(0x01);
+        if (!endpoint) {
+            qCWarning(dcZigbeeRemotes()) << "Device claims to be an Insta remote but does not provide endpoint 1";
+            return false;
+        }
+
+        createThing(instaThingClassId, networkUuid, node, endpoint);
+
+        // Nothing to be done here... The device does not support battery level updates and will send all the commands
+        // to the coordinator unconditionally, no need to bind any clusters...
```
t-mon, line 68: Are you sure about this, does it also after a complete factory reset? No leftover configurations and bindings in the device from previous networks? Is there a possibility to totally factory reset the device and test if it works?

zigbeeremotes/integrationpluginzigbeeremotes.cpp
```diff
@@ -0,0 +1,288 @@
...
+            qCDebug(dcZigbeeRemotes()) << "Scenes command received:" << command << groupId << sceneId;
+            thing->emitEvent(instaPressedEventTypeId, {Param(instaPressedEventButtonNameParamTypeId, QString::number(sceneId))});
+        });
+
+
+        // The device also supports setting saturation, color and color temperature. However, it's quite funky
+        // actually get there on the device and that mode seems to be only enabled if there are bindings to
+        // actual lamps. We'd need to fake a lamp and bind it to it.
+        // (Short-press on and off nultiple times to select one of the bound lights, then the scenes buttons will control color/temp/sat)
+
```
t-mon, line 160: There should be no need to fake a lamp, the binding theoretically should work if you bind them to the coordinator.

zigbeeremotes/integrationpluginzigbeeremotes.json
```diff
@@ -0,0 +1,75 @@
+{
+    "name": "ZigbeeRemotes",
+    "displayName": "Zigbee remotes",
+    "id": "d0184b49-6e6d-42fe-92e1-e2ede67ca7df",
+    "vendors": [
+        {
+            "name": "nymea",
+            "displayName": "nymea GmbH",
+            "id": "2062d64d-3232-433c-88bc-0d33c0ba2ba6",
```
t-mon, line 9: Since this is an actual custom node implementation from a manufacturer you know, we should go with the that vendor? I understand the reason for the generic remotes plugin, but in this case giving the vendor makes sense I think.

t-mon (changes requested):

t-mon (approved):

t-mon (approved):

Issue #528, opened 2022-01-14
SMA: Manual setup not working because mac address parameter doesn't have a default value


mzanetti:

t-mon:
has been fixed in #529

PR #529, opened 2022-01-17
SMA: Add default value for sma sunny webbox thing param #528

t-mon:

### DIFF
--- a/sma/integrationpluginsma.json
+++ b/sma/integrationpluginsma.json
@@ -29,7 +29,8 @@
                             "displayName": "hardware address",
                             "type": "QString",
                             "inputType": "TextLine",
-                            "readOnly": true
+                            "readOnly": true,
+                            "defaultValue": "00:00:00:00:00:00"
                         }
                     ],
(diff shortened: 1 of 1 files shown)


sma/integrationpluginsma.json
```diff
@@ -29,7 +29,8 @@
                             "displayName": "hardware address",
                             "type": "QString",
                             "inputType": "TextLine",
-                            "readOnly": true
+                            "readOnly": true,
+                            "defaultValue": "00:00:00:00:00:00"
```
mzanetti, line 33: I believe the setupThing needs a check if it's the null mac address in order for the manual setup to work (without having tried - just reading the code).
In that context I'd probably go with "" as default value...

t-mon: The input type should be MacAddress and not TextLine I think, that's why I use the zero mac and also: it's read only...

The auto reconnect / rediscovery is planed for the next release, so that might change some things...

mzanetti (approved):

PR #543, opened 2022-02-17
Keba: Add series capability check and fetch product information during setup

t-mon:

### DIFF
--- a/keba/README.md
+++ b/keba/README.md
@@ -6,13 +6,14 @@ This plugin allows to control Keba KeContact EV-Charging stations.
 
 * KeContact Wallbox
 	* P20 (certain models)
-	* P30 (certain models)
+	* P30
+		* c-series
+		* x-series
 	* BMW (certain models)
-	* [Keba Deutschland Edition](https://a.storyblok.com/f/40131/x/fc59dc7bf7/datenblatt_deutschland_edition.pdf)
-	* [Keba Green Edition](https://keba.com/download/x/7ec5b6e96f/121218_dbde.pdf)
+	* [Keba Deutschland Edition](https://a.storyblok.com/f/40131/x/fc59dc7bf7/datenblatt_deutschland_edition.pdf) (DE440)
+(by March 2022)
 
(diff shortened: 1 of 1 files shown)


keba/kebaproductinfo.h
```diff
@@ -0,0 +1,132 @@
...
+#include <QObject>
+#include <QString>
+
+class KebaProductInfo
+{
+    Q_GADGET
+public:
+    enum Connector {
+        ConnectorSocket,
+        ConnectorCabel
```
mzanetti, line 43: Typo, Cable, but I guess "Wire" would be the most common term in our this context

keba/kebaproductinfo.h
```diff
@@ -0,0 +1,132 @@
...
+
+    enum ConnectorCurrent {
+        Current13A = 1,
+        Current16A = 2,
+        Current20A = 3,
+        Current32A = 4
+    };
+    Q_ENUM(ConnectorCurrent)
+
+    enum Cabel {
```
mzanetti, line 62: Wire, or Cable

keba/integrationpluginkeba.cpp
```diff
@@ -182,6 +185,33 @@ void IntegrationPluginKeba::setupThing(ThingSetupInfo *info)
                 return;
             }
 
+            // Parse the product code and check if the model actually supports the communication
+            // Only series X and C support UDP/Modbus
+            if (productInformation.isValid()) {
```
mzanetti, line 190: I wonder if we shouldn't bail out instead of pass when the product string can't be parsed... Do you think there is any situation where the product string cannot be processed but all the rest would still work?

mzanetti (changes requested):

keba/kebaproductinfo.h
t-mon: _A wire is a single conductor(material most commonly being copper or aluminium) while cable is two or more insulated wires wrapped in one jacket._

I think cable is more accuarte...

PR #546, opened 2022-03-16
New plugin - Daikin Airco

loosrob:
first version of plugin, must functionality present


daikinairco/integrationplugindaikinairco.json
```diff
@@ -0,0 +1,131 @@
...
+                            "type": "QString",
+                            "writable": true,
+                            "possibleValues": [
+                                "Automatic",
+                                "Cooling",
+                                "Heating",
+                                "Dehumidifier",
+                                "Fan"
+                            ],
+                            "defaultValue": "Automatic"
```
mzanetti, line 121: The thermostat interface has states for "heatingOn" and "coolingOn". Any chance to get those out of the this API?

daikinairco/integrationplugindaikinairco.json
```diff
@@ -0,0 +1,131 @@
...
+                            "displayName": "Connected",
+                            "displayNameEvent": "Connected changed",
+                            "defaultValue": false,
+                            "type": "bool",
+                            "cached": false
+                        },
+                        {
+                            "id": "bd0ff439-6ab5-414a-ba6e-9ecef0f7c3f4",
+                            "name": "url",
+                            "displayName": "Device Address",
```
mzanetti, line 49: this shouldn't be a state IMO but rather stored in the pluginStorage if needed. But actually I think this should be just a volatile variable in the plugin and refreshed with findIps() every time it is unknown

daikinairco/integrationplugindaikinairco.py
```diff
@@ -0,0 +1,387 @@
...
+                homeTemp = float(splitItem[1])
+                airco.setStateValue(aircoTemperatureStateTypeId, homeTemp)
+            elif splitItem[0] == 'hhum':
+                homeHumidity = float(splitItem[1])
+                airco.setStateValue(aircoHumidityStateTypeId, homeHumidity)
+            elif splitItem[0] == 'otemp':
+                outTemp = float(splitItem[1])
+                airco.setStateValue(aircoOutsideTemperatureStateTypeId, outTemp)
+    else:
+        airco.setStateValue(aircoConnectedStateTypeId, False)
```
mzanetti, line 199: If I read this correctly it would set the device as offline here but then still proceed with the next call below which then will, by all odds, fail too.

So I think it should just exit the polling here, even though it doesn't do much harm either if it stays like this. Just a thought.

daikinairco/integrationplugindaikinairco.py
```diff
@@ -0,0 +1,387 @@
...
+        logger.log("timer interval @ setupThing", pollTimer.interval)
+    else:
+        logger.log("Timer already exists @ setupThing")
+    
+    info.finish(nymea.ThingErrorNoError)
+    return
+
+def pollAirco(info):
+    global pollFrequency
+    deviceIp = info.stateValue(aircoUrlStateTypeId)
```
mzanetti, line 174: As noted in the json, a state is not the right place to memorize the IP. Note that every state change is sent out through the network to all clients which results in unneeded traffic.

In fact, what if the DHCP lease time expires and the device is rebooted and gets a new IP?
AFAICT you only call findIPs() during discovery and in setupThing(). This would require to restart nymea in order to make it pick up the new ip. Wouldn't it make more sense to just use findIps() here instead to always have a fresh one?
That, of course, would result in frequent unneeded discoveries which isn't super nice either. So I suppose the most correct solution would be to check here if the device is marked as disconnected. If so, run findIps here. If it shows up as connected, do the calls (by memorizing the Ips in a local variable). If the call then fails, it would set the device as offline again and the next poll cycle would start with the discovery again.

mzanetti (changes requested):

daikinairco/integrationplugindaikinairco.json
loosrob: Can't be obtained via the API, but I'll calculate them based on mode, temperature & target temperature -- should be doable

daikinairco/integrationplugindaikinairco.json
loosrob: I've used states in previous plugins as well, as that seemed easiest. Happy to use variable instead, but I would need some help with that, as I can't really say how that would work with multiple devices --> variable per device?

daikinairco/integrationplugindaikinairco.py
loosrob: I'll try to find the IP again, as suggested below, and exit the polling.

daikinairco/integrationplugindaikinairco.json
mzanetti: Is that how the device behaves? Always heating or cooling if the current temperature is unequal the target temperature? If you can calculate them close to what the device is actually doing, then great, of your guess would be far off from reality, better leaving them away.

daikinairco/integrationplugindaikinairco.json
mzanetti: Ah, sorry, that must have slipped my reviews then.
So, you already used that, for example in the neatobotvac plugin, for instance accountsMap. You can have a `ipMap` or similar where you save the ip to with `ipMap[thing.id] = ip` and read it with `ip = ipMap[thing.id]`.

daikinairco/integrationplugindaikinairco.json
loosrob: Some quick tests seem to indicate this is close to what the device is doing; but I'll monitor more long term.

daikinairco/integrationplugindaikinairco.json
loosrob: I see, thanks!

loosrob: I addressed this comment in the latest commit (tested OK).

daikinairco/meta.json
```diff
@@ -0,0 +1,12 @@
...
+    "title": "Daikin Airco",
+    "tagline": "Discover and control Daikin airconditioning devices",
+    "icon": "daikin.jpg",
+    "stability": "testing",
+    "offline": true,
+    "technologies": [
+        "network"
+    ],
+    "categories": [
+    ]
```
mzanetti, line 11: add `appliance`and `heating`

mzanetti (changes requested):

daikinairco/integrationplugindaikinairco.json
mzanetti: Are we ok with this then?

loosrob: Yes, seems OK after the last changes I made

mzanetti:
on amd64, installing dependencies fails with

```ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
anyio 3.3.2 requires idna>=2.8, but you have idna 2.5 which is incompatible.
Successfully installed certifi-2020.12.5 chardet-4.0.0 distro-1.5.0 httplib2-0.19.0 idna-2.5 ifaddr-0.1.7 lazr.restfulclient-0.14.3 lazr.uri-1.0.5 oauthlib-3.1.0 pbr-5.5.1 pyparsing-2.4.7 requests-2.25.1 requests-oauthlib-1.3.0 setuptools-54.1.1 six-1.15.0 testresources-2.0.1 urllib3-1.26.3 wadllib-1.3.5
```

debian/control
```diff
@@ -163,6 +163,22 @@ Description: nymea.io plugin for commandlauncher
  This package will install the nymea.io plugin for commandlauncher
 
 
+Package: nymea-plugin-daikinairco
+Architecture: any
+Depends: ${shlibs:Depends},
+         ${misc:Depends},
+         nymea-plugins-translations,
```
mzanetti, line 170: please remove this. the nymea-plugins-translations package will be removed with 1.1

debian/control
```diff
@@ -163,6 +163,22 @@ Description: nymea.io plugin for commandlauncher
...
+         python3-pip,
+Description: nymea.io plugin for Daikin airconditioning units
+ The nymea daemon is a plugin based IoT (Internet of Things) server. The
+ server works like a translator for devices, things and services and
+ allows them to interact.
+ With the powerful rule engine you are able to connect any device available
+ in the system and create individual scenes and behaviors for your environment.
+ .
+ This package will install the nymea.io plugin for Daikin airconditioning units.
```
mzanetti, line 179: Also in 1.1 we'll update all the descriptions to remove all of this copy/paste text and we'll just have:

```
Description: nymea.io plugin for Daikin airconditioning units
 This package contains the nymea integration plugin for Daikin airconditioning units.
```

mzanetti (changes requested):

mzanetti (approved):

PR #548, opened 2022-03-22
New Plugin: Meross (via Meross cloud)

mzanetti:


mzanetti:
The meross cloud only allows one call to the device per 10 seconds or the device will be blocked from the cloud and needs to be reset and configured again. This PR would need to be tuned to prevent that from happening.
However, having much better results with the cloud-less variant so focusing on that.

PR #554, opened 2022-04-11
New plugin: Meross

mzanetti:


meross/integrationpluginmeross.cpp
```diff
@@ -0,0 +1,366 @@
...
+}
+
+IntegrationPluginMeross::~IntegrationPluginMeross()
+{
+}
+
+void IntegrationPluginMeross::discoverThings(ThingDiscoveryInfo *info)
+{
+    NetworkDeviceDiscoveryReply *reply = hardwareManager()->networkDeviceDiscovery()->discover();
+    connect(reply, &NetworkDeviceDiscoveryReply::finished, reply, &NetworkDeviceDiscoveryReply::deleteLater);
```
t-mon, line 58: No need to do this, the reply is doing this on it's own.

meross/integrationpluginmeross.cpp
```diff
@@ -0,0 +1,366 @@
...
+    connect(reply, &NetworkDeviceDiscoveryReply::finished, info, [info, reply, this](){
+        foreach (const NetworkDeviceInfo &deviceInfo, reply->networkDeviceInfos()) {
+            qCDebug(dcMeross) << "Discovery result" << deviceInfo;
+            if (deviceInfo.hostName().startsWith("Meross_Smart_Plug")) {
```
t-mon, line 62: I always compare such things case in-sensitive, seen wired stuff over the years...

meross/integrationpluginmeross.cpp
```diff
@@ -0,0 +1,366 @@
...
+            foreach (Thing *thing, myThings()) {
+                pollDevice5s(thing);
+            }
+        });
+    }
+    if (!m_pluginTimer10s) {
+        m_pluginTimer10s = hardwareManager()->pluginTimerManager()->registerTimer(10);
+        connect(m_pluginTimer10s, &PluginTimer::timeout, this, [this](){
+            foreach (Thing *thing, myThings()) {
+                pollDevice60s(thing);
```
t-mon, line 172: 10s timer calls poll device 60s?

meross/integrationpluginmeross.cpp
```diff
@@ -0,0 +1,366 @@
...
+
+    QUrl url;
+    url.setScheme("http");
+    url.setHost(m_deviceMonitors.value(thing)->networkDeviceInfo().address().toString());
+    url.setPath("/config");
+    QNetworkRequest request(url);
+    request.setHeader(QNetworkRequest::ContentTypeHeader, "application/json");
+
+    qCDebug(dcMeross) << "Requesting with key" << key << qUtf8Printable(QJsonDocument::fromVariant(data).toJson());
+
```
t-mon, line 360: The idea of the monitor is: if the monitor says the thing is nor reachable, there is no point in sending a request to the device.

meross/integrationpluginmeross.cpp
```diff
@@ -0,0 +1,366 @@
...
+void IntegrationPluginMeross::setupThing(ThingSetupInfo *info)
+{
+    pluginStorage()->beginGroup(info->thing()->id().toString());
+    m_keys.insert(info->thing(), pluginStorage()->value("key").toByteArray());
+    pluginStorage()->endGroup();
+
+    if (!m_deviceMonitors.contains(info->thing())) {
+        NetworkDeviceMonitor *monitor = hardwareManager()->networkDeviceDiscovery()->registerMonitor(MacAddress(info->thing()->paramValue(plugThingMacAddressParamTypeId).toString()));
+        m_deviceMonitors.insert(info->thing(), monitor);
```
t-mon, line 157: You can connect to the reachable changed signal: once reachable again, you can make a initial call and then continue polling. That would prevent flooding of warnings if not reachable.

t-mon (changes requested):

meross/integrationpluginmeross.cpp
mzanetti: yes, which is why the SMA plugin picked up this plug :D
But fair enough. changed it

mzanetti: yes, which is why the SMA plugin picked up this plug :D
But fair enough. changed it

meross/integrationpluginmeross.cpp
mzanetti: Hmm... So I've inhibited the polling when it's not reachable now, but extra code to stop the pluginTimer and relaunch it when the device comes reachable again seems to messy for the little gain... I'm thinking though... perhaps it would make sense to rethink the plugintimer... @fetzerch also asked todday about a dynamic interval changing option in the pluginTimer api. So if we'd provide a pause/resume for that too, it would be nice to do something like this:

```
void setupThing(ThingSetupInfo *info) {
  NetworkDeviceMonitor* monitor = hardwareManager()->networkDeviceMonitor()->registerMonitor(...);
  PluginTimer *timer = hardwareManager()->pluginTimerManager()->registerTimer(5);
  connect(monitor, &NetworkDeviceMonitor::reachableChanged, timer, &PluginTimer::setActive);
  connect(timer, &PluginTimer::triggered, info->thing(), [](){
    pollThing();
  } 
}
```

mzanetti: Hmm... So I've inhibited the polling when it's not reachable now, but extra code to stop the pluginTimer and relaunch it when the device comes reachable again seems to messy for the little gain... I'm thinking though... perhaps it would make sense to rethink the plugintimer... @fetzerch also asked todday about a dynamic interval changing option in the pluginTimer api. So if we'd provide a pause/resume for that too, it would be nice to do something like this:

```
void setupThing(ThingSetupInfo *info) {
  NetworkDeviceMonitor* monitor = hardwareManager()->networkDeviceMonitor()->registerMonitor(...);
  PluginTimer *timer = hardwareManager()->pluginTimerManager()->registerTimer(5);
  connect(monitor, &NetworkDeviceMonitor::reachableChanged, timer, &PluginTimer::setActive);
  connect(timer, &PluginTimer::triggered, info->thing(), [](){
    pollThing();
  } 
}
```

PR #555, opened 2022-04-13
TcpCommander: Add TCP server broadcast functionality to send data to all connected clients

t-mon:

### DIFF
--- a/tcpcommander/integrationplugintcpcommander.cpp
+++ b/tcpcommander/integrationplugintcpcommander.cpp
@@ -54,9 +54,9 @@ void IntegrationPluginTcpCommander::setupThing(ThingSetupInfo *info)
             // In case of a reconfigure, make sure we reconnect
             tcpSocket->disconnectFromHost();
         }
+
         connect(tcpSocket, &QTcpSocket::stateChanged, thing, [=](QAbstractSocket::SocketState state){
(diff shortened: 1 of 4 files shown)


mzanetti (changes requested):
Mostly ok, but the change to the connected logic is not good.

The current code intentionally only indicates being connected if at least one client is connected, given TCP allows us to know that.

If you really want to change that behavior (I probably wouldn't but don't have a strong opinion on it either), you'd also have to change the code in `onTcpServerConnectionCountChanged` to not set it to disconnected when the last client disconnects.

t-mon:
> Mostly ok, but the change to the connected logic is not good.

Besides the fact that a server has no connected state, the state should reflect if the server is listening, not if there is a client connected. The client counter shows you how many are connected, but without this change, everything looks broken if no client is connected.

mzanetti:
> > Mostly ok, but the change to the connected logic is not good.
> 
> Besides the fact that a server has no connected state, the state should reflect if the server is listening, not if there is a client connected. The client counter shows you how many are connected, but without this change, everything looks broken if no client is connected.

Fair enough. but then also the clientDisconnected slot needs to be fixed.

mzanetti (approved):
lgtm now

PR #565, opened 2022-05-20
goe: Update discovery and add API V2 support

t-mon:


t-mon:
One note: it is not ready to be landed. 
We still have to test this with hardware V1 / V2 to make sure API V1 is still working as expected. 
Unfortunately on hardware V3 (introducing API V2) the APIV1 is not working the same way as on hardware V1/V2...
But it is ready to review.

goecharger/integrationplugingoecharger.cpp
```diff
@@ -199,52 +229,23 @@ void IntegrationPluginGoECharger::executeAction(ThingActionInfo *info)
...
         uint maxChargingCurrent = action.paramValue(goeHomeMaxChargingCurrentActionMaxChargingCurrentParamTypeId).toUInt();
         qCDebug(dcGoECharger()) << "Setting max charging current to" << maxChargingCurrent << "A";
-        // FIXME: check if we can use amx since it is better for pv charging, not all version seen implement amx
-        // Maybe check if the user sets it or a automatism
-        QString configuration = QString("amp=%1").arg(maxChargingCurrent);
-        sendActionRequest(thing, info, configuration);
+        executeSetChargingCurrent(info, maxChargingCurrent);
         return;
-    } else if (action.actionTypeId() == goeHomeAbsoluteMaxAmpereActionTypeId) {
+    }/* else if (action.actionTypeId() == goeHomeAbsoluteMaxAmpereActionTypeId) {
```
mzanetti, line 243: state isn't writable any more. can probably be deleted

goecharger/integrationplugingoecharger.cpp
```diff
@@ -483,73 +568,420 @@ void IntegrationPluginGoECharger::sendActionRequest(Thing *thing, ThingActionInf
...
+            update(thing, statusMap);
+
+            // From now on we listen to topics
+            connect(channel, &MqttChannel::publishReceived, thing, [=](MqttChannel* channel, const QString &topic, const QByteArray &payload){
+                QString propertyKey = topic.split("/").last();
+                QString propertyValue = QString::fromUtf8(payload);
+                // Well...no better idea for now to keep the APIs / parsing methods
+                // compatible trought different APIs and protocols
+                QString statusMapJsonString = QString("{\"%1\":%2}").arg(propertyKey).arg(propertyValue);
+                QJsonDocument jsonDoc = QJsonDocument::fromJson(statusMapJsonString.toUtf8());
```
mzanetti, line 926: How is this different than just directly creating a `QMap<QString, QVariant>{{propertyKey, propertyValue}}` ? 
This looks like it would easily break parsing, for instance if a non-numeric or empty value is received. If going through the json parser is really required, shouldn't the %2 have "" around it too to not fail the parser? No error check needed in case it fails?

goecharger/integrationplugingoecharger.cpp
```diff
@@ -624,22 +1056,22 @@ void IntegrationPluginGoECharger::setupMqttChannel(ThingSetupInfo *info, const Q
...
                     } else {
-                        qCDebug(dcGoECharger()) << "Configured successfully MQTT server" << thing << channel->password();
+                        qCDebug(dcGoECharger()) << "Configured successfully MQTT server" << thing << channel->username();
                     }
 
-                    // Enable MQTT
-                    QNetworkRequest request = buildConfigurationRequest(address, QString("mce=1"));
-                    qCDebug(dcGoECharger()) << "Enable custom mqtt server on" << thing << request.url().toString();
+                    // Password
+                    QNetworkRequest request = buildConfigurationRequest(thing, QUrlQuery(QString("mck=%1").arg(channel->password())));
```
mzanetti, line 1068: using this QUrlQuery ctor you probably want to `toPercentageEncoding()` the password. This will break if the password contains a & or a =

mzanetti: otherwise use `Query::setQueryItem("mck", password)` which would do the right thing.

goecharger/integrationplugingoecharger.cpp
```diff
@@ -654,52 +1086,224 @@ void IntegrationPluginGoECharger::setupMqttChannel(ThingSetupInfo *info, const Q
...
+
+        // Build the mqtt url
+        QUrl mqttUrl;
+        mqttUrl.setScheme("mqtt");
+        mqttUrl.setHost(channel->serverAddress().toString());
+        mqttUrl.setPort(channel->serverPort());
+        mqttUrl.setUserName(channel->username());
+        mqttUrl.setPassword(channel->password());
+
+        // The query item must be JSON encoded, meaning: strings need quouts... for whatever reason...
```
mzanetti, line 1183: quotes :)

goecharger/integrationplugingoecharger.cpp
```diff
@@ -654,52 +1086,224 @@ void IntegrationPluginGoECharger::setupMqttChannel(ThingSetupInfo *info, const Q
...
+            }
+
+            qCDebug(dcGoECharger()) << "Configuration of MQTT for" << thing << "finished successfully";
+            // From now on we listen to topics
+            connect(channel, &MqttChannel::publishReceived, thing, [=](MqttChannel* channel, const QString &topic, const QByteArray &payload){
+                QString propertyKey = topic.split("/").last();
+                QString propertyValue = QString::fromUtf8(payload);
+                // Well...no better idea for now to keep the APIs / parsing methods
+                // compatible trought different APIs and protocols
+                QString statusMapJsonString = QString("{\"%1\":%2}").arg(propertyKey).arg(propertyValue);
```
mzanetti, line 1221: duplicate code? Given it's quite funky code it would probably make sense to only have it in a single place?

goecharger/integrationplugingoecharger.json
```diff
@@ -277,6 +291,15 @@
...
                             "defaultValue": 0.00
                         },
+                        {
+                            "id": "28f59f96-4b30-4606-9a04-80c82abc346b",
+                            "name": "frequency",
+                            "displayName": "Frequency",
+                            "displayNameEvent": "Frequency changed",
+                            "type": "double",
+                            "unit": "Hertz",
+                            "defaultValue": 50.0
```
mzanetti, line 301: disable caching I'd say... also a lot of other states seem to be caching while it's not needed (and causes a lot of disk I/O)

mzanetti (changes requested):

goecharger/integrationplugingoecharger.cpp
t-mon: yeah I know how wired this is, but have a look here:
https://github.com/goecharger/go-eCharger-API-v2/blob/main/http-en.md#setting-values

It actually mixes JSON and URL standards for whatever reason...maybe to keep parser code within the controller simple...without this one I actually get a JSON error in return to a HTTP GET request

goecharger/integrationplugingoecharger.cpp
t-mon: I know...we might watch how it develops, as of now point to the actually issue


https://github.com/goecharger/go-eCharger-API-v2/issues/73

t-mon:
Note: this branch will be replaced with #587 since there is nothing in common with API V1 on HW V3 and API V1 on HW V1/2. Let's keep things separated.

PR #566, opened 2022-05-21
somfytahoma: Use local connection "Somfy Developer Mode"

fetzerch:
This PR changes the somfytahoma plugin to directly connect to Somfy Tahoma Gateways without having to rely on the Somfy Tahoma cloud connection.

In order to activate this local API, users need to enable the so called 'Developer Mode' on their gateways. See https://developer.somfy.com/developer-mode. Existing plugin users will have to reconfigure the plugin.

The old cloud based access had several disadvantages:
- Somfy changed the login method that caused the old version of the plugin to disconnect after a few days.
- Somfy started to disconnect and block 3rd party applications that polled device updates too fast.

The local access now offers various advantages:
- We don't rely on an internet connection any more (except for initial access to the box) which should drastically improve stability and performance of the integration.
- The 'Developer Mode' is officially supported, documented and officially grants access for 3rd Party applications. See https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode.

Additional changes in this PR:
- Add support for Smoke Detector.

---
Known Issues:

- [ ] Event polling started to fail after a couple of days, but there's no usable debug output. I've improved the error logging now.
- [ ] Access tokens are valid only for 6 months. Need to add logic to refresh the token.

Fixed Issues:

- [x] Unregistering the token on removal of the gateway fails. Need to investigate how the API is meant to be used.
- [x] Plugin never uses the IP it gets through Zeroconf. This was caused by an incorrect filter.

---

### DIFF
--- a/somfytahoma/integrationpluginsomfytahoma.json
+++ b/somfytahoma/integrationpluginsomfytahoma.json
@@ -12,8 +12,7 @@
                     "id": "fedd72b8-547d-4e4f-b73e-71344a8ba0c1",
                     "name": "tahoma",
                     "displayName": "Tahoma Account",
-                    "createMethods": ["user"],
-                    "setupMethod": "userandpassword",
+                    "createMethods": ["auto"],
                     "interfaces": ["account"],
                     "stateTypes": [
                         {
@@ -46,14 +45,23 @@
                     "id": "6c09e0b9-f0cc-4dea-9994-9e039eff78f1",
                     "name": "gateway",
                     "displayName": "Tahoma Gateway",
-                    "createMethods": ["auto"],
+                    "createMethods": ["discovery"],
+                    "setupMethod": "userandpassword",
                     "interfaces": ["gateway"],
                     "paramTypes": [
                         {
                             "id": "e321a7d6-6dcb-4a37-baf1-c7008f2d5bdb",
                             "displayName": "Gateway Id",
                             "name": "gatewayId",
-                            "type": "QString"
+                            "type": "QString",
+                            "readOnly": true
+                        },
+                        {
+                            "id": "30b73244-e5bb-4c00-9332-702a60c03420",
+                            "displayName": "Gateway pin",
+                            "name": "gatewayPin",
+                            "type": "QString",
+                            "readOnly": true
                         }
                     ],
                     "stateTypes": [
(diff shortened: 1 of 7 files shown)


somfytahoma/integrationpluginsomfytahoma.json
```diff
@@ -46,14 +45,37 @@
...
+                        {
+                            "id": "cd9673c8-37a3-4692-afae-592097aba6e9",
+                            "displayName": "API version",
+                            "name": "apiVersion",
+                            "type": "QString",
+                            "readOnly": true
+                        },
+                        {
+                            "id": "36639f5d-26cb-4666-a148-13485bb1ccdc",
+                            "displayName": "Firmware version",
```
mzanetti, line 75: FWIW, there is a `update` interface available which would have a state for the current device firmware version along with "updateState" etc. Not sure if it's possible to trigger a firmware update via the API. If so, this param would make more sense being a state following the interface.
Generally I for one try to use as few "params" as needed to identify and setup a device as params cannot really be modified from the plugin side of things but are meant to be set by the user (or the discovery) for setting something up.
Anyhow, now that this would be an actual issue, I'd approve the PR nevertheless.

somfytahoma/somfytahomarequests.cpp
```diff
@@ -38,7 +38,8 @@
 #include "extern-plugininfo.h"
 
 
-static const QString somfyTahomaUrl = QStringLiteral("https://tahomalink.com/enduser-mobile-web/enduserAPI");
+static const QString somfyTahomaWebUrl = QStringLiteral("https://ha101-1.overkiz.com/enduser-mobile-web/enduserAPI");
```
mzanetti, line 41: Is this really the URL suggested by Somfy? Kinda looks like something that would be returned from a load balancer instead of the actual entry point url.

mzanetti (approved):
Looks good. 2 inline comments, but I'm really just asking. If you say this is ok, no need to change anything.

somfytahoma/somfytahomarequests.cpp
fetzerch: Yep, this is what the documentation suggests: https://github.com/Somfy-Developer/Somfy-TaHoma-Developer-Mode#api-authentication.

Somfy seems to by the IoT stack from a company called OverKiz. This is why that name pops up here and there.

somfytahoma/integrationpluginsomfytahoma.json
fetzerch: Removed firmware and api version since they're not needed to identify a device.

PR #586, opened 2022-07-20
SMA: Improve Sunny WebBox discovery

t-mon:

### DIFF
--- a/sma/sunnywebboxdiscovery.cpp
+++ b/sma/sunnywebboxdiscovery.cpp
@@ -1,0 +1,163 @@
+/* * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * * *
+*
+* Copyright 2013 - 2022, nymea GmbH
+* Contact: contact@nymea.io
+*
+* This file is part of nymea.
+* This project including source code and documentation is protected by
+* copyright law, and remains the property of nymea GmbH. All rights, including
+* reproduction, publication, editing and translation, are reserved. The use of
+* this project is subject to the terms of a license agreement to be concluded
+* with nymea GmbH in accordance with the terms of use of nymea GmbH, available
+* under https://nymea.io/license
+*
+* GNU Lesser General Public License Usage
+* Alternatively, this project may be redistributed and/or modified under the
+* terms of the GNU Lesser General Public License as published by the Free
+* Software Foundation; version 3. This project is distributed in the hope that
+* it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+* warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+* Lesser General Public License for more details.
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with this project. If not, see <https://www.gnu.org/licenses/>.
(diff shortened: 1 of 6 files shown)


sma/sunnywebboxdiscovery.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    // When the network discovery has finished, we process the rest and give some time to finish the pending replies
+    connect(m_discoveryReply, &NetworkDeviceDiscoveryReply::finished, this, [=](){
+        // The network device discovery is done
+        m_discoveredNetworkDeviceInfos = m_discoveryReply->networkDeviceInfos();
+        m_discoveryReply = nullptr;
+
+        // Check if all network device infos have been verified
+        foreach (const NetworkDeviceInfo &networkDeviceInfo, m_discoveredNetworkDeviceInfos) {
+            if (m_verifiedNetworkDeviceInfos.contains(networkDeviceInfo))
```
mzanetti, line 79: You seem to never insert something into m_verifiedNetworkDeviceInfos, which, if I'm reading it correctly will cause all the initially checked devices to be checked twice.

But be aware, if you just add `m_verifiedNetworkDeviceInfos.append(networkDeviceInfo)` in `checkNetworkDevice()` after the successful verification, it will still be racy as chances are, it's pending the network reply still... But I think inserting into m_verifiedNetworkDeviceInfos right at the top of checkNetworkDevice() could work...

sma/sunnywebboxdiscovery.cpp
```diff
@@ -0,0 +1,163 @@
...
+void SunnyWebBoxDiscovery::startDiscovery()
+{
+    // Clean up
+    m_discoveryResults.clear();
+    m_verifiedNetworkDeviceInfos.clear();
+    m_gracePeriodTimer.stop();
+
+    m_startDateTime = QDateTime::currentDateTime();
+
+    qCInfo(dcSma()) << "Discovery: SunnyWebBox: Start discovering the network...";
```
mzanetti, line 60: Network discovery starting...

Otherwise the user may start running around in his home :-P

sma/sunnywebboxdiscovery.cpp
```diff
@@ -0,0 +1,163 @@
...
+             return;
+         }
+    });
+}
+
+void SunnyWebBoxDiscovery::cleanupPendingReplies()
+{
+    foreach (QNetworkReply *reply, m_pendingReplies) {
+        m_pendingReplies.removeAll(reply);
+        reply->abort();
```
mzanetti, line 151: Hmm... I think this abort() call here would still cause the above lambda to trigger, whcih contains does the m_pendingReplies.removeAll(reply) again and then prints the HTTP error. Probably not an actual issue, but perhaps take a minute to check if the cleanup is really happening properly and its what you want.

mzanetti (changes requested):

sma/sunnywebboxdiscovery.cpp
t-mon: right...that's how it is ment to be used, just forgott due to lack of testing possibilities...good catch!

mzanetti (approved):

PR #599, opened 2022-08-25
myStrom: Improve discovery mechanism

mzanetti:

### DIFF
--- a/mystrom/integrationpluginmystrom.cpp
+++ b/mystrom/integrationpluginmystrom.cpp
@@ -59,16 +59,52 @@ void IntegrationPluginMyStrom::init()
 
 void IntegrationPluginMyStrom::discoverThings(ThingDiscoveryInfo *info)
 {
+    QList<QNetworkReply*> *pendingReplies = new QList<QNetworkReply*>();
+    connect(info, &ThingDiscoveryInfo::finished, this, [pendingReplies](){
+        delete pendingReplies;
+    });
+
     foreach (const ZeroConfServiceEntry &entry, m_zeroConf->serviceEntries()) {
         qCDebug(dcMyStrom()) << "Found myStrom device:" << entry;
         if (entry.protocol() != QAbstractSocket::IPv4Protocol) {
             continue;
         }
-        ThingDescriptor descriptor(switchThingClassId, entry.name(), entry.hostAddress().toString());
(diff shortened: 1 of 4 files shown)


mystrom/integrationpluginmystrom.cpp
```diff
@@ -59,16 +59,48 @@ void IntegrationPluginMyStrom::init()
...
+                return;
+            }
+            QByteArray data = reply->readAll();
+            QJsonParseError error;
+            QJsonDocument jsonDoc = QJsonDocument::fromJson(data, &error);
+            if (error.error != QJsonParseError::NoError) {
+                finishDiscoveryReply(reply, info, pendingReplies);
+                return;
+            }
+            qCDebug(dcMyStrom) << "Info response:" << qUtf8Printable(jsonDoc.toJson());
```
t-mon, line 93: Do you want to keep this?

mystrom/integrationpluginmystrom.cpp
```diff
@@ -59,16 +59,48 @@ void IntegrationPluginMyStrom::init()
 
 void IntegrationPluginMyStrom::discoverThings(ThingDiscoveryInfo *info)
 {
+    QList<QNetworkReply*> *pendingReplies = new QList<QNetworkReply*>();
+    connect(info, &ThingDiscoveryInfo::finished, this, [pendingReplies](){
+        delete pendingReplies;
+    });
+
```
t-mon, line 66: If there is no zeroconf available on if there arn't any service entries in the network, you never finish the discovery until timeout. You could fallback to network device discovery although. If that's to much overhead now, you should:

```
if (m_zeroConf->serviceEntries().isEmpty()) {
    info->finish(Thing::ThingErrorNoError);
    return;
}
```

t-mon (changes requested):

mystrom/integrationpluginmystrom.cpp
mzanetti: nice catch. fixed

mystrom/integrationpluginmystrom.cpp
mzanetti: Yes, in case the discovery fails for a user because different models have different type ids (i.e. if a french plug has ID 108 or something) we'd see that in users bug reports right away if debug logs are enabled and wouldn't have to ask back to somehow fetch that info from the user.

t-mon (approved):

PR #653, opened 2023-01-23
go-e: Add possibility to control charging phases

mzanetti:

### DIFF
--- a/goecharger/integrationplugingoecharger.json
+++ b/goecharger/integrationplugingoecharger.json
@@ -322,6 +322,18 @@
                             "defaultValue": 1
                         },
                         {
+                            "id": "db0af9a7-08fd-4224-b071-c89e11ae8c47",
+                            "name": "desiredPhaseCount",
+                            "displayName": "Desired number of charging phases",
+                            "displayNameAction": "Set desired number of charging phases",
+                            "type": "uint",
+                            "possibleValues": [1, 3],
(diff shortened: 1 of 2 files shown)


goecharger/integrationplugingoecharger.json
```diff
@@ -321,6 +321,18 @@
...
                             "defaultValue": 1
                         },
+                        {
+                            "id": "db0af9a7-08fd-4224-b071-c89e11ae8c47",
+                            "name": "desiredPhaseCount",
+                            "displayName": "Desired number of charging phases",
+                            "displayNameAction": "Set desired number of charging phases",
+                            "type": "uint",
+                            "possibleValues": [1, 3],
+                            "defaultValue": 1,
```
t-mon, line 331: Since this feature is not working/existing for goe- API v1 (Hardware V1), I'd suggest to either: create 2 thing classes depending on the hardware / API version, one with this feature, one without

or

as an alternative, the action would always fail (as now with the wrong error code `ThingErrorActionTypeNotFound`). Not supported or something woul make more sense, since th actiontype definitily exists, but the action would not be supported by the hardware.

In that case, default should be 3 phases and you should return a propper error code in the APIv1 section. By default, hardware V1 supports 3 phase charging.

t-mon (changes requested):

PR #668, opened 2023-02-17
Shelly: Add support for the Plus Plug S

loosrob:

### DIFF
--- a/shelly/integrationpluginshelly.cpp
+++ b/shelly/integrationpluginshelly.cpp
@@ -299,7 +299,7 @@ void IntegrationPluginShelly::discoverThings(ThingDiscoveryInfo *info)
         } else if (info->thingClassId() == shelly2ThingClassId) {
             namePattern = QRegExp("^shellyswitch-[0-9A-Z]+$");
         } else if (info->thingClassId() == shelly25ThingClassId) {
-            namePattern = QRegExp("^shellyswitch25-[0-9A-Z]+$");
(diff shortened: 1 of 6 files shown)


shelly/integrationpluginshelly.cpp
```diff
@@ -1690,6 +1705,45 @@ void IntegrationPluginShelly::setupGen2(ThingSetupInfo *info)
...
+                    config.insert("initial_state", defaultState);
+                    QVariantMap params;
+                    params.insert("id", 0);
+                    params.insert("config", config);
+
+                    ShellyRpcReply *reply2 = client->sendRequest("Switch.SetConfig", params);
+                    connect(reply2, &ShellyRpcReply::finished, info, [info](ShellyRpcReply::Status status, const QVariantMap &/*response*/){
+                        if (status != ShellyRpcReply::StatusSuccess) {
+                            qCWarning(dcShelly) << "Error during shelly setup";
+                            info->finish(Thing::ThingErrorHardwareFailure);
```
mzanetti, line 1725: the issue with this is that earlier, in line 1659 we've already called info->finish() with success. At that point the info object will be destroyed and this slot will never be called. The options are either to not access info any more here and just ignore any failures, or, don't call finish() above but only call it with success here. I'd prefer the second option.

Now I see that the same mistake is already in the plus2pm branch and I've fixed it. So please rebase this one again and adjust in here as I did for the plus2pm.

mzanetti (changes requested):

PR #671, opened 2023-02-21
FEMS: Added support for the FEMS API

DerStoecki:
With this Plugin it is  possible to connect to the FEMS via REST.
(Easy to add additional datapoints if some additional data is needed etc)

NOTE: Tested with OpenEMS-Edge not a FEMS system (but  its core is openems-edge so logic would be the same)

- [✓] Make sure the pull request's title is of format "Plugin name: Add support for xyz" or "New plugin: Plugin name"

- [✓] Did you test the changes on hardware, if not (e.g. absence of required hardware), please mention a person to confirm it has been tested.

- [✓] Did you update the plugin's README.md accordingly?

- [✓] Did you update translations (`cd builddir && make lupdate`)?


mzanetti:
Thanks for the contribution. Please accept the CLA to allow us to actually use this.

mzanetti:
Thanks for signing the CLA.

So, to begin with, in order to allow integrating this plugin into the nymea-plugins reporistory, the debian packaging structure for the standalone packaging would need to be removed and instead hooked into the main plugin dpkg. For that, please just remove the entire fems/debian folder and add just the package into the main repositories debian/control file, as well as providing the according debian/nymea-plugins-fenecon.install.in.

Generally we name the packages according to the vendor name, so in this case it would probably be nymea-plugin-fenecon instead of nymea-plugin-fems. That would allow extending the plugin also for other devices by fenecon, as well as ease up searching for the user as I think a user would rather search for fenecon instead of fems.

mzanetti:
I've also looked into this device recently and it seems that it would also support modbus TCP. Now I'm totally fine with using the REST API instead of Modbus but just out of curiosity, what was your rationale for choosing the REST api over modbus?

DerStoecki:
> I've also looked into this device recently and it seems that it would also support modbus TCP. Now I'm totally fine with using the REST API instead of Modbus but just out of curiosity, what was your rationale for choosing the REST api over modbus?

Hi sorry for the delayed answer, I've been to the OpenEMS Hackathon and I'll answer with the github app. Thanks for the suggestions, I'll edit the packages next week on monday/Tuesday depends on my schedule.
The reason for the usage of the rest api instead of the Modbus api has 2-3 reasons.
The logic of the openems edge /fems are so called channeladresses. Those addresses are a concatenation of the configured openems device id and 'channel' ID. Those will remain the same /are more easy to grasp than the Modbus api.
Next reason is
The Modbus channel may change from fems to fems depending on the configuration of the fems therefore its pretty hard to say 1 Modbus address might stay the same and means the same thing. But tbh I don't know much about the fems product itself and more about the openems stack.
And last but not least by using the rest api you can easily connect to an openems edge instead of a fems (compatible to both) or can set up your own openems and connect to nymea via the rest api(as long as your are using device id's such as meter0, ess0 and so on)
And if you want to expand the code and connect to... Let's say an evcs... You can have a look into the openems repo - > see the channels - >implement those very quick into nymea. As long as the ID stays the same (usually evcs0, meter0, ess0) you can always be assured to get the correct data. By using the Modbus api (if you have a look into the openems repo - > controller.modbus.api) you can see that it depends on the order of components what Modbus address certain data points  will be assigned to.

(to be fair the 'sum' data points will stay at the same address)

However, REST with openems is way more powerful since you can read any 'channel' data point from an openems component.

fems/integrationpluginfems.json
```diff
@@ -0,0 +1,477 @@
...
+                    "createMethods": ["user"],
+                    "interfaces": ["connectable"],
+                    "paramTypes": [
+                    {
+                        "id": "297e85cc-e2aa-41fe-a27a-c3904b2e7f70",
+                        "name": "address",
+                        "displayName": "Host address",
+                        "type": "QString",
+                        "inputType": "IPv4Address",
+                        "defaultValue": "192.168.158.2"
```
mzanetti, line 25: Hmm... This is something we're quite strictly avoiding in nymea... IPs will change over time and this will start to fail. 

We'd need to keep a unique device ID (serial number, network MAC address, ...) instead and use that instead in the params and resolve the IP dynamically. For that nymea offers helper tools which make it quite easy to do so: Preferrably openEMS would announce itself on mDNS, then you could use the ZeroConfDiscovery. Otherwise you could use the NetworkDeviceDiscovery.

fems/integrationpluginfems.json
```diff
@@ -0,0 +1,477 @@
...
+                        "id": "73814aef-571d-4a42-ba76-5751f128d86b",
+                        "name": "Port",
+                        "displayName": "Port",
+                        "type": "QString",
+                        "defaultValue": "80"
+                    },
+                    {
+
+                       "id": "ede7a93e-36a1-4873-8ff8-d71e621baa6f",
+                       "name": "edge",
```
mzanetti, line 37: afaics this is only used to determine whether authentication is done in the url or in the Authorization header. Could this be detected in one way or another. Looking at it from a user point of view, this is very unclear what the user should do here during the setup.

fems/integrationpluginfems.json
```diff
@@ -0,0 +1,477 @@
...
+            "name": "fems",
+            "displayName": "Fems",
+            "id": "c241f100-35fc-498f-a4a0-61b7ee274470",
+            "thingClasses": [
+                {
+                    "name": "connection",
+                    "displayName": "Fems System",
+                    "id": "2649f540-8f17-462d-93f7-79aa9cfc3734",
+                    "setupMethod": "userandpassword",
+                    "createMethods": ["user"],
```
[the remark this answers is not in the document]
DerStoecki: You could argue you would be able to but, afaik the FEMS is most commonly run on a cloud/different server and not in the local network.

fems/integrationpluginfems.json
DerStoecki: Yes i've don this approach with a modbusplugins, however it's difficult to rely on a MAC Address for a FEMS system. Afaik the OpenEMS/FEMS can run on a Virtual Machine or Server not in your local Network -> therefor it's difficult/if not impossible (to my knowledge at least) to store a non IP Address identifier

fems/integrationpluginfems.cpp
```diff
@@ -0,0 +1,1405 @@
...
+                            << connection->address().toString();
+        return;
+    }
+    // 3 things, parent, and battery as well as meter but only parent connection
+    // nec. GET Meter and Battery ID Later and update param info
+    Thing *connectionThing = m_femsConnections.value(connection);
+    if (!connectionThing)
+        return;
+    qCDebug(dcFems()) << "Updating States";
+    switch (connectionSwitch) {
```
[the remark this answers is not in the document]
DerStoecki: Yes i totally agree regarding FemsConnections, however the requirement I got stated to only allow 1 FEMS system.

---
I agree, I can remove the switch() it's not necessary. And OpenEMS can easily handle those few REST calls.

fems/integrationpluginfems.json
DerStoecki: Ahhh yes i get it. I looked at this the following way:
Since it doesn't matter logically if you are connecting to a "FEMS" or a "OpenEMS-Edge" system I wanted to support both.
But i can see the confusion regarding endusers...i could either remove the support for OpenEMS-Edge systems all together or remove the authentication via URL / dedicated FEMS connection. Whatever you prefer.

mzanetti: Right... So there are 2 options I suppose:
1) Remove upstream OpenEMS support from here and make the best possible user experience for Fenecon FEMS users.
2) Make it an OpenEMS plugin with the best possible user experience for OpenEMS users and remove the name FEMS.

I think option 1 is the way to go here, would you agree? In that sense, we can drop this. Now there's one thing that confuses me. If "useEdge" is false, this plugin will use http://user:pass@host as authentication. When useEdge is true, it will do the authentication in the Authorize header instead. I have tested it with a Fenecon FEMS system and had to use the Authorize header as in-url auth wasn't working. So this seems to be the wrong way round in here or what am I missing?

fems/integrationpluginfems.json
mzanetti: Looking at the Fenecon website, it seems the system will always run locally on the so called "FEMS Box" component of the system. Am I wrong there?

DerStoecki: Ahhh I see, well in this case it would be possible to implement to look for a FEMS System in a local network.
It'll take some time to implement the auto-discovery though, (busy schedule) but i'll try my best :)

fems/integrationpluginfems.json
DerStoecki: Interesting...I implemented the System as mentioned in the documentation, but I only tested this with an OpenEMS-Edge System. Sorry for the inconvenience

I think I'll remove the URL Authentication, and just use the authorization header instead.
This is a quick fix.

fenecon/README.md
```diff
@@ -0,0 +1,100 @@
+# Fenecon
+## FEMS
+--------------------------------
+
+Nymea plugin to read some data from a FEMS. This can be extended in the future easily by adding different "Natures" of OpenEMS etc.
```
mzanetti, line 5: The README.md file will be used to generate help pages for the end user. Please only keep information relevant to the end user in here. No developer gibberish a end user won't understand. If you consider it worth adding a README for developers (such as the status bits, links to the rest api etc), please put that in another file, e.g. README-dev.md

This should really just list the supported devices, and any details requried for the user on how to set it up. e.g. if the user needs to explicitly enable the REST api in the FEMS system, then this would be the right place to explain that.

fenecon/README.md
```diff
@@ -0,0 +1,100 @@
...
+fems plugin
+
+## Configuration
+
+Configuration parameters are
+
+* Name
+* Host Address
+* Port
+* If the device is an Edge (e.g. not fems but openems directly)
```
mzanetti, line 99: Outdated. but also not required. The user will see what's needed during the setup when the app asks for it.

fenecon/integrationpluginfenecon.cpp
```diff
@@ -0,0 +1,1386 @@
...
+}
+// JSON of FEMS contains value behind "value"
+QString IntegrationPluginFenecon::getValueOfRequestedData(QJsonDocument *json) {
+  return json->toVariant().toMap().value(DATA_ACCESS_STRING_FEMS).toString();
+}
+
+void IntegrationPluginFenecon::addValueToThing(Thing *parentThing,
+                                               ThingClassId identifier,
+                                               StateTypeId stateName,
+                                               const QVariant *value,
```
mzanetti, line 1272: why a QVariant *? If you use `const QVariant &value` instead, you can delete a lot of "new" and "delete" lines :)

fenecon/README.md
DerStoecki: ahhh i see, alright i'll update the readme/readme-dev :)

fems/integrationpluginfems.json
DerStoecki: I'm sorry but i don't quite get it...
How do you setup your thing via devicediscovery and after selecting the device you ask the user after their username and pwd for the fems connection?
I'll push my current code and wait for a response (thank you in advance)

mzanetti: * discoveryThings() is called by the system. Find the device (e.g. via ZeroConf, NetworkDiscovery , UPnP, whatever) and return the thingdescriptor to it (no login at this point)
* startPairing() is called by the system. Most likely you don't need to do anything in here, you can set a pretty printed text which credentials the user should enter here
* confirmPairing() is called by the system: in here you get the user/password. Do a test login and call info->finish(NoError) if successful or an appropriate error otherwise and store the login credentials in the pluginStorage()
* setupThing() is called by the system. Load the credentials from the pluginStorage() and do the regular setup stuff.

DerStoecki: things can be quite simple nice :D thank you its working with my local openems-edge. perfect!

fems/integrationpluginfems.cpp
```diff
@@ -0,0 +1,1405 @@
...
+                .filterByThingClassId(femsstatusThingClassId)
+                .length() < 1 &&
+                !this->statusCreated) {
+            ThingDescriptor descriptor(femsstatusThingClassId, "FEMS Status", QString(),
+                                       connectionThing->id());
+            ParamList params;
+            params.append(Param(femsstatusThingIdParamTypeId, "Status"));
+            descriptor.setParams(params);
+            emit autoThingsAppeared(ThingDescriptors() << descriptor);
+            this->statusCreated = true;
```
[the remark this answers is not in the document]
DerStoecki: "Also this has the problem with multiple FEMS systems and a single statusCreated variable instance." 

We have the requirement to only allow 1 FEMS system with 1 Battery and 1 meter/inverter, since its an unusual case to use multiple FEMS Systems in your Home/Commercial system.

UseCases for multiple FEMS System: connecting multiple FEMS System and let them connect via EdgeToEdge Communication.
This happens only in large Commercial Projects (afaik).

Since the FEMS is an EnergyManagementSystem itself, that connects to multiple peripheral devices, you can read every data from the FEMS via REST as long as you know the device Ids and Channels. And therefore do not need multiple FEMS systems.

Anyway our usecase is only "1 FEMS with only 1 device of selected types available on runtime".

------

"That said. I'd like to challenge the need for this femsstatusThingClassId. If every connection has this status thing, IMO it would make more sense to have just one ThingClass with 2 states. A separate thing will just clutter the UX."
^However I changed that, i deleted the fems stateclass id and moved that to the connection / parent.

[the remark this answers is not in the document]
mzanetti: Well, it's fine for you to connect just 1 but in order for a plugin to be merged to upstream nymea, it needs to work in all circumstances, that implies handling this properly. After all it's really not that hard to fix this :)

Anyhow, looks like you did fix it, so we're good. I see the statusCreated variable is still around but not used any more. I suppose it can be dropped now.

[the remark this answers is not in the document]
DerStoecki: "Anyhow, looks like you did fix it, so we're good. I see the statusCreated variable is still around but not used any more. I suppose it can be dropped now."

whoops that was an oversight :D sorry for that, ill fix that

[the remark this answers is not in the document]
DerStoecki: " it needs to work in all circumstances"
Oh in this case we would/you would need to refactor quite a bit:

1. Get the EdgeConfig via RPC call (almost same path as REST slightly different, but same authentication and port..on default)
2. Read every config entry -> check service pid -> you get the device service pid + device id and can map internally what device is enabled and what kind of device it is (chp, heatpump, asymmetric/symmetric meter, battery, evcs etc etc) 
then map to an internal device (nymea: EnergyStorage, smartmeter etc)/save the device id (in case the default id changed)
3. If you want to only support base "natures" (e.g. base interfaces like your smartmeter, battery etc) its okayish to do that (regarding time + effort to do that)
4. if you want more details you would need to support every expansion of those base natures.
most data you could get via the sum/ess0 
but if you want to support more batteries / the whole OpenEMS / FEMS system ...thats quite a bit of work.

If you would want to support those cases I would propose to either implemented this as a BETA version...or put this into a feature branch of yours.

-----
This would be easy for me if this was in a framework i know quite well / programming language i am at least intermediate in.

But if it is alright with you I'd prefer to keep this implementation the way it is and either implement this as a BETA version, or if you want to support the FEMS system, you could use this as a template / starting point if you want :)
If "working under all circumstances"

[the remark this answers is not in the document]
mzanetti: When I say "it needs to work in all circumstances" I don't mean it needs to support all possible variations of OpenEMS or be 100% feature complete. That means it should not have any architectural issues that would prevent adding 2 FEMS systems and fixing that would require a rework of the plugin that would break existing setups.

So I think we're ok to only support the Fenecon FEMS and I'm also ok if it isn't 100% feature complete as long as it can be extended later without a complete refactoring.

Without having tried it yet, if you removed the 2 places I discovered in the previous review, I think this doesn't look to bad now. In fact, thanks a lot for the great job in implementing all my suggestions and I think we're quite close now in having this PR approved.

PR #706, opened 2023-06-19
Keba: Fix type mismatch warning on setState for phaseCount

t-mon:


mzanetti:
Hmm... Just reading it I don't think that'll fix the issue... I mean, yes, it will get rid of the warning because a invalid QVariant will be converted to 0 on a toUint() but I suppose the actual issue in a different place, perhaps even the core or the app, in that it feeds the API with a invalid QVariant in the first place... Can we reproduce it somehow?

PR #746, opened 2024-10-28
Shelly: Improve debug logging when offline

martinlukas84:
When shelly integration can't join the multicast group it begins spaming the log each 5s. With this commit it log the joining problem each 5s for the first Minute, after that each 10min.

### DIFF
--- a/shelly/integrationpluginshelly.cpp
+++ b/shelly/integrationpluginshelly.cpp
@@ -768,30 +768,30 @@ void IntegrationPluginShelly::joinMulticastGroup()
 {
     if (m_coap->joinMulticastGroup()) {
         qCInfo(dcShelly()) << "Joined CoIoT multicast group";
-        m_multicastRetryCnt = 0;
+        m_multicastWarningPrintCount = 0;
     } else {
-        uint retryTime = 0;
+        uint mod = m_multicastWarningPrintCount % 120;
 
         // FIXME: It would probably be better to monitor the network interfaces and re-join if necessary
-        if (m_multicastRetryCnt < 12) {
+        if (m_multicastWarningPrintCount < 12) {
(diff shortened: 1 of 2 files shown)


shelly/integrationpluginshelly.cpp
```diff
@@ -768,12 +768,23 @@ void IntegrationPluginShelly::joinMulticastGroup()
...
+        m_multicastRetryCnt = 0;
     } else {
-        qCWarning(dcShelly()) << "Failed to join CoIoT multicast group. Retrying in 5 seconds...";
+        uint retryTime = 0;
+
         // FIXME: It would probably be better to monitor the network interfaces and re-join if necessary
-        QTimer::singleShot(5000, m_coap, [this](){
+        if (m_multicastRetryCnt < 12) {
+            qCWarning(dcShelly()) << "Failed to join CoIoT multicast group. Retrying in 5 seconds...";
+            retryTime = 5000;
```
t-mon, line 778: I would suggest to always retry every 5 seconds, only print the warning every 10 minutes after one minute grace period.
Since we cannot tell yet when a multicast group might be available (yet), we should continue with the joining retries.

shelly/integrationpluginshelly.h
```diff
@@ -94,6 +94,7 @@ private slots:
     PluginTimer *m_reconfigureTimer = nullptr;
 
     Coap *m_coap = nullptr;
+    uint m_multicastRetryCnt = 0;
```
t-mon, line 97: Suggestion: rename the the member to `m_multicastWarningPrintCount` since the count should only be related to the warning printing, not the retry itself.

t-mon (changes requested):

t-mon (approved):

